PR #1, opened 2015-12-04
Fix a build error on Darwin.

ufosky:


gribozavr:
I'm sorry, but what is the error you are seeing?  This code compiles fine for me on OS X.

ufosky:
The default shell bash will produce this error:

```
Unknown option: 1
Usage: head [-options] <url>...
    -m <method>   use method for the request (default is 'HEAD')
    -f            make request even if head believes method is illegal
    -b <base>     Use the specified URL as base
    -t <timeout>  Set timeout value
    -i <time>     Set the If-Modified-Since header on the request
    -c <conttype> use this content-type for POST, PUT, CHECKIN
    -a            Use text mode for content I/O
    -p <proxyurl> use this as a proxy
    -P            don't load proxy settings from environment
    -H <header>   send this HTTP header (you can specify several)

    -u            Display method and URL before any response
    -U            Display request headers (implies -u)
    -s            Display response status code
    -S            Display response status chain
    -e            Display response headers
    -d            Do not display content
    -o <format>   Process HTML content in various ways

    -v            Show program version
    -h            Print this message

    -x            Extra debugging output
```

ufosky:
This is the evaluated command:

``` shell
sh -c "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld -v 2>&1 | head -1"
```

gribozavr:
What command do you use to build swift?

gribozavr:
And which version of Xcode do you have?  You need the latest beta: https://developer.apple.com/xcode/download/ Please xcode-select it after installing.

ufosky:
system:  OS X 10.0.5
shell: bash 3.2
xcode: 7.1
build command: ./swift/utils/build-script -t

if I use zsh,the command is OK.

gribozavr:
Could you post the output of `sh -c "which head"`?

ufosky:
Oh,it's not the system head.
/Applications/XAMPP/bin/head

PR #3, opened 2016-02-02
Introduce ns_error_domain attribute.

milseman:
ns_error_domain can be used by, e.g. NS_ERROR_ENUM, in order to
identify a global declaration representing the domain constant.

Introduces the attribute, Sema handling, diagnostics, and test case.

### DIFF
--- a/include/clang/Basic/Attr.td
+++ b/include/clang/Basic/Attr.td
@@ -1080,6 +1080,12 @@ def ObjCBridgeRelated : InheritableAttr {
   let Documentation = [Undocumented];
 }
 
+def NSErrorDomain : Attr {
+  let Spellings = [GNU<"ns_error_domain">];
+  let Documentation = [Undocumented];
+  let Args = [IdentifierArgument<"ErrorDomain">];
+}
+
 def NSReturnsRetained : InheritableAttr {
   let Spellings = [GNU<"ns_returns_retained">];
 //  let Subjects = SubjectList<[ObjCMethod, ObjCProperty, Function]>;
--- a/lib/Sema/SemaDeclAttr.cpp
+++ b/lib/Sema/SemaDeclAttr.cpp
@@ -4057,6 +4057,38 @@ static void handleObjCRequiresSuperAttr(Sema &S, Decl *D,
                                         attr.getAttributeSpellingListIndex()));
 }
 
+static void handleNSErrorDomain(Sema &S, Decl *D, const AttributeList &Attr) {
+  if (!isa<TagDecl>(D)) {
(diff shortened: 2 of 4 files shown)


include/clang/Basic/Attr.td
```diff
@@ -1080,6 +1080,12 @@ def ObjCBridgeRelated : InheritableAttr {
   let Documentation = [Undocumented];
 }
 
+def NSErrorDomain : Attr {
+  let Spellings = [GNU<"ns_error_domain">];
+  let Documentation = [Undocumented];
```
DougGregor, line 1085: Please add some documentation here, otherwise it won't get documented _anywhere_.

lib/Sema/SemaDeclAttr.cpp
```diff
@@ -4057,6 +4057,38 @@ static void handleObjCRequiresSuperAttr(Sema &S, Decl *D,
...
+  // Verify that the identifier is a valid decl in the C decl namespace
+  LookupResult lookupResult(S, DeclarationName(identLoc->Ident),
+                            SourceLocation(),
+                            Sema::LookupNameKind::LookupOrdinaryName);
+  if (!S.LookupName(lookupResult, S.TUScope)) {
+    S.Diag(identLoc->Loc, diag::err_nserrordomain_invalid_decl)
+        << identLoc->Ident;
+    return;
+  }
+
```
DougGregor, line 4086: Perhaps you should be checking that it's actually a global-scope variable?

DougGregor, line 4086: ... in which case you could record the declaration in the attribute, rather than just the identifier, so clients don't need to look it up themselves.

DougGregor:
Note that this pull request should be against upstream-with-swift, not stable. We can cherry-pick into stable after accepting the pull request.

lib/Sema/SemaDeclAttr.cpp
milseman, line 4086: Yeah, I wanted to leave the actual Decl in the attribute, but it didn't seem like I could without modifying the tablegen backend for Attributes. I'm certainly open to doing that, but not sure if it's worth it. Right now, only FunctionDecl and IdentifierInfo are supported, it seems, in ClangAttrEmitter.cpp's "createArgument".

milseman:
I will make a new PR for the right branch, as I think I have to close this one to change what it merges into

lib/Sema/SemaDeclAttr.cpp
DougGregor, line 4086: Okay, that's fine

PR #5, opened 2016-02-02
Introduce ns_error_domain attribute.

milseman:
ns_error_domain can be used by, e.g. NS_ERROR_ENUM, in order to
identify a global declaration representing the domain constant.

Introduces the attribute, Sema handling, diagnostics, and test case.

### DIFF
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7533,6 +7533,14 @@ def err_nsconsumed_attribute_mismatch : Error<
 def err_nsreturns_retained_attribute_mismatch : Error<
   "overriding method has mismatched ns_returns_%select{not_retained|retained}0"
   " attributes">;
+
+def err_nserrordomain_not_tagdecl : Error<
+  "ns_error_domain attribute only valid on "
+  "%select{enums, structs, and unions|enums, structs, unions, and classes}0">;
+def err_nserrordomain_invalid_decl : Error<
+  "domain argument %0 does not refer to global constant">;
+def err_nserrordomain_requires_identifier : Error<
+  "domain argument must be an identifier">;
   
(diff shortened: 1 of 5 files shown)


milseman:
@swift-ci Please test

DougGregor:
swift-ci isn't doing branches now. The change LGTM; go ahead and merge, then cherry-pick into stable and stable-next.

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7533,6 +7533,13 @@ def err_nsconsumed_attribute_mismatch : Error<
 def err_nsreturns_retained_attribute_mismatch : Error<
   "overriding method has mismatched ns_returns_%select{not_retained|retained}0"
   " attributes">;
+
+def err_nserrordomain_not_tagdecl : Error<
+  "ns_error_domain attribute only valid on enum/struct/union/class">;
```
jrose-apple: I'm pretty sure it's only valid on an enum. If we wanted to allow it on anything else, it would be typedefs for bare integer types.

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7533,6 +7533,13 @@ def err_nsconsumed_attribute_mismatch : Error<
...
+def err_nserrordomain_invalid_decl : Error<
+  "domain argument %0 not valid top-level declaration">;
```
jrose-apple: How about "domain argument %0 does not refer to a global constant"? Even if we don't actually check all of those things, it guides the user in the right direction.

milseman: Will do!

milseman:
Alright, I updated with Jordan's suggested diagnostic improvements. Will merge now

PR #7, opened 2016-02-12
Fix autotools build

jroelofs:


jrose-apple:
The autotools build of LLVM/Clang isn't supported in any Swift branches (and has been removed from upstream LLVM).

jroelofs:
Ah, I didn't catch that. Are there plans to merge cbeineman's
upstream patch to remove the autotools build to this branch?

Jon

jrose-apple:
No, it's just that it's languished and never been tested with Swift. At this point the 2.2 Clang branch is pretty much frozen, but we should probably at least merge Chris's commit to the 3.0 branch, now that you mention it.

gribozavr:
We will merge it as a part of the stable merge later.

PR #8, opened 2016-02-14
Enable Objective-C fixed enums feature for Swift

hpux735:
This patch enables CF_ENUM and CF_OPTIONS on non-Darwin targets

cc/ @tienex

### DIFF
--- a/lib/Lex/PPMacroExpansion.cpp
+++ b/lib/Lex/PPMacroExpansion.cpp
@@ -1040,6 +1040,12 @@ static void ComputeDATE_TIME(SourceLocation &DATELoc, SourceLocation &TIMELoc,
 }
 
 
+static bool hasModuleFeature(StringRef Feature, const LangOptions &LangOpts) {
+    return std::find(LangOpts.ModuleFeatures.begin(),
+                     LangOpts.ModuleFeatures.end(),
+                     Feature) != LangOpts.ModuleFeatures.end();
+}
+
 /// HasFeature - Return true if we recognize and implement the feature
 /// specified by the identifier as a standard language feature.
 static bool HasFeature(const Preprocessor &PP, const IdentifierInfo *II) {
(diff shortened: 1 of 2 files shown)


gribozavr:
This change needs tests...

gribozavr:
CC @jrose-apple

hpux735:
Oops, sorry.  I made this PR first.  Here is the test: https://github.com/apple/swift/pull/1311

jrose-apple:
Interesting. My other thought was that we'd just add this as an extension to C in upstream Clang, if we can sell that to the Clang maintainers. @DougGregor, what do you think?

jrose-apple:
If we _do_ do this, we want a Clang-side test too, though. I think there's a flag to add arbitrary module features.

DougGregor:
I think we should start the discussion on the Clang development mailing list (http://lists.llvm.org/mailman/listinfo/cfe-dev) to add this feature as an extension to C, because that would be the best overall result.

compnerd:
@DougGregor, @jrose-apple I find your idea rather intriguing and think that it
is a great path forward.  I am somewhat concerned about compatibility though (as
it pertains to Linux).  Would the following small twist on your idea be
palatable?

Introduce `-fapple-extensions` as a flag to enable additional extensions to help
improve swift interoperability on non-Darwin environments.  This would allow us
to retain GCC compatible behaviour on Linux as well as enable them as an opt-in
when desired (e.g. when working with swift) as it already requires clang to be
in use.

Also, Im happy to move the discussion to cfe-dev to get these extensions enabled.

DougGregor:
I'd like to avoid adding something like `-fapple-extensions` if we can, because having Swift import a different variant of C than is normally used is likely to cause integration problems down the road. From our perspective, it would be far better to allow this extremely-useful extension in C to keep things consistent. I suspect that this extension will be well-received, but we won't know until someone opens up the discussion on cfe-dev.

compnerd:
Started a thread on cfe-dev under the subject of "Enabling strong enum extensions under C".

hpux735:
Here's a link to @compnerd 's thread http://lists.llvm.org/pipermail/cfe-dev/2016-February/047364.html for posterity and convenience.

PR #9, opened 2016-02-18
fix MSVC compatibility

vitalyster:
- add missing CINDEX_LINKAGE to exported functions
- eliminate C2276 MSVC++ error on static member reference


DougGregor:
I cherry-picked the appropriate change from upstream for CINDEX_LINKAGE into swift-2.2-branch. The SourceMgrAdapter fix actually needs to go into the upstream-with-swift branch, and then we'll cherry-pick it appropriately. The branching scheme is described at

  https://swift.org/contributing/#llvm-and-swift

but it's silly to ask for a new PR for a one-liner like this, so I went ahead and committed/cherry-picked in the appropriate places. Thanks!

PR #18, opened 2016-06-22
Allow iOS and tvOS version numbers with 2-digit major version numbers.

bob-wilson:
rdar://problem/26921601


bob-wilson:
@swift-ci Please test

PR #33, opened 2016-10-24
[AST] Remove superfluous docblock change

modocache:
Upstream Clang documents the `Message` parameter as simply `\param`, while Swift's fork of Clang documents it as `\param[out]`. Although `\param[out]` may be more correct, the change to the documentation should be made in upstream Clang, not Swift Clang.

Furthermore, upstream Clang does not appear to have a strong convention as to the use of `\param[out]`: of 4616 instances of `\param`, only 49 are denoted as `\param[out]`.

---

This reduces the diff between upstream and Swift Clang, from 156 modified files to 155.

### DIFF
--- a/include/clang/AST/DeclBase.h
+++ b/include/clang/AST/DeclBase.h
@@ -605,7 +605,7 @@ public:
   /// the given declaration (e.g., preferring 'unavailable' to
   /// 'deprecated').
   ///
-  /// \param[out] Message If non-NULL and the result is not \c
+  /// \param Message If non-NULL and the result is not \c
   /// AR_Available, will be set to a (possibly empty) message
   /// describing why the declaration has not been introduced, is
   /// deprecated, or is unavailable.


jrose-apple:
This seems all right, but `stable` is the wrong branch for it. Changes to LLVM and Clang for Swift purposes should go into `upstream-with-swift`, so they get picked up in later rebranches as well.

modocache:
@jrose-apple Thanks for the tip -- I was definitely confused as to which branch was best. Done!

jrose-apple (approved):
Feel free to merge without testing.

modocache:
I don't have write access to this repository, so I can't merge myself.

Should commit access to apple/swift also come with commit access to apple/swift-clang, @shahmishal?

shahmishal:
@modocache Sorry, I am not sure. This sounds like question from code-owners@swift.org. Thanks!

modocache:
OK! I sent an email just now.

PR #34, opened 2016-10-24
[Sema] Display an objc_subclassing_restricted error for Objective-C implementation declarations when appropriate.

hyp:


jrose-apple:
Thanks for the patch! However, since this is a feature implemented in upstream Clang, the patch should really go there. You can submit it at reviews.llvm.org (and assign me as a reviewer there).

hyp:
It looks like ObjCSubclassingRestrictedAttr isn't upstreamed.

hyp:
Should I work on upstreaming this attribute?

jrose-apple:
Ah, I didn't realize that. Okay, then this makes sense. I wonder if we should actually just warn on the `@implementation` all the time, though. Is there any reason to be warning on the `@interface` at all?

hyp:
Warning on `@implementation` all the time should work, but I think because of dynamic nature of Objective-C it's better to try to warn on the `@interface` first. This way the diagnostic will show up even if the Objective-C code contains no `@implementation` for an `@interface`. But it's not really that important, and I will change the patch if you think that we should use only the `@implementation` diagnostic.

jrose-apple:
The way I see it, "the dynamic nature of ObjC" doesn't really apply here; an `@interface` on its own doesn't do anything at run-time except declare symbols, and those symbols won't be defined without an `@implementation`.

My thought was that if someone else's code is doing this, you don't really need to see the error in your header file, only when compiling their library. That does open the door to people trying to conditionally include it, though, so that they can use this like the `public`/`open` distinction in Swift. That's probably not something we want to introduce on a whim; at the very least it would need more discussion among ObjC stakeholders at Apple.

Okay, let's do it this way. Thanks, Alex! (And thanks for helping me understand this.)

jrose-apple (approved):

PR #36, opened 2016-11-08
Fix use-of-temporary with StringRef in code coverage

jrose-apple:
Caught by adding a similar check to StringRef as r283798 did for ArrayRef. Reviewed by @vedantk as https://reviews.llvm.org/D26317.


jrose-apple:
@swift-ci Please test

PR #37, opened 2016-11-10
[APINotes] Update for upstream changes to llvm::BitstreamReader.

jrose-apple:
See https://reviews.llvm.org/D26219. rdar://problem/29078044


jrose-apple:
Tests passed locally, and we don't have PR testing for upstream-with-swift. Merging.

PR #39, opened 2016-11-11
[APINotes] Add a 'SwiftImportAsAccessors' entry for properties.

jrose-apple:
If true, Swift will import the property as a getter and setter (or just a getter for read-only properties), rather than as a Swift 'var' property. At least, it will once this lands and the Swift side gets implemented.

Part of rdar://problem/28455962

### DIFF
--- a/include/clang/APINotes/Types.h
+++ b/include/clang/APINotes/Types.h
@@ -336,8 +336,13 @@ public:
 
 /// Describes API notes data for an Objective-C property.
 class ObjCPropertyInfo : public VariableInfo {
+  unsigned SwiftImportAsAccessorsSpecified : 1;
+  unsigned SwiftImportAsAccessors : 1;
+
 public:
-  ObjCPropertyInfo() : VariableInfo() { }
+  ObjCPropertyInfo()
+      : VariableInfo(), SwiftImportAsAccessorsSpecified(false),
+        SwiftImportAsAccessors(false) {}
 
   /// Merge class-wide information into the given property.
   friend ObjCPropertyInfo &operator|=(ObjCPropertyInfo &lhs,
@@ -349,6 +354,32 @@ public:
       }
     }
 
+    return lhs;
+  }
+
+  Optional<bool> getSwiftImportAsAccessors() const {
+    if (SwiftImportAsAccessorsSpecified)
+      return SwiftImportAsAccessors;
+    return None;
+  }
+  void setSwiftImportAsAccessors(Optional<bool> value) {
+    if (value.hasValue()) {
+      SwiftImportAsAccessorsSpecified = true;
+      SwiftImportAsAccessors = value.getValue();
+    } else {
+      SwiftImportAsAccessorsSpecified = false;
+      SwiftImportAsAccessors = false;
+    }
+  }
+
+  friend ObjCPropertyInfo &operator|=(ObjCPropertyInfo &lhs,
+                                      const ObjCPropertyInfo &rhs) {
+    lhs |= static_cast<const VariableInfo &>(rhs);
+    if (!lhs.SwiftImportAsAccessorsSpecified &&
+        rhs.SwiftImportAsAccessorsSpecified) {
+      lhs.SwiftImportAsAccessorsSpecified = true;
+      lhs.SwiftImportAsAccessors = rhs.SwiftImportAsAccessors;
+    }
     return lhs;
   }
 };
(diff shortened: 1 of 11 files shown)


jrose-apple:
@DougGregor, since this is my first API notes patch, would you mind reviewing?

include/clang/APINotes/Types.h
```diff
@@ -336,8 +336,11 @@ class VariableInfo : public CommonEntityInfo {
 
 /// Describes API notes data for an Objective-C property.
 class ObjCPropertyInfo : public VariableInfo {
+  unsigned SwiftImportAsAccessorsSpecified : 1;
+  unsigned SwiftImportAsAccessors : 1;
+
 public:
-  ObjCPropertyInfo() : VariableInfo() { }
+  ObjCPropertyInfo() : VariableInfo(), SwiftImportAsAccessorsSpecified(false) {}
```
DougGregor: I recommend also initializing `SwiftImportAsAccessors`, even if it seems silly.

include/clang/APINotes/Types.h
```diff
@@ -351,6 +354,31 @@ class ObjCPropertyInfo : public VariableInfo {
...
+    if (SwiftImportAsAccessorsSpecified)
+      return SwiftImportAsAccessors;
+    return None;
+  }
+  void setSwiftImportAsAccessors(Optional<bool> value) {
+    if (value.hasValue()) {
+      SwiftImportAsAccessorsSpecified = true;
+      SwiftImportAsAccessors = value.getValue();
+    } else {
+      SwiftImportAsAccessorsSpecified = false;
```
DougGregor, line 370: Same here; let's just set `SwiftImportAsAccessors = false;`.

lib/APINotes/APINotesWriter.cpp
```diff
@@ -606,11 +606,17 @@ namespace {
...
     }
 
     void emitUnversionedInfo(raw_ostream &out, const ObjCPropertyInfo &info) {
       emitVariableInfo(out, info);
+      uint8_t flags = 0;
+      if (Optional<bool> value = info.getSwiftImportAsAccessors()) {
+        flags |= 1 << 0;
+        flags |= value.getValue() << 1;
+      }
+      out << flags;
```
DougGregor, line 619: We generally don't stream values out to the `raw_ostream`;  please use the endian writer, e.g.,

```
endian::Writer<little> writer(out);
writer.write<uint18_t>(flags);
```

lib/APINotes/APINotesYAMLCompiler.cpp
```diff
@@ -784,6 +786,8 @@ namespace {
         pInfo.SwiftName = prop.SwiftName;
         if (prop.Nullability)
           pInfo.setNullabilityAudited(*prop.Nullability);
+        if (prop.SwiftImportAsAccessors)
+          pInfo.setSwiftImportAsAccessors(*prop.SwiftImportAsAccessors);
```
DougGregor, line 790: This file also has a translation in the _other_ direction that you'll need to do, e.g., from the `ObjCPropertyInfo` to YAML. It's used to make sure everything round-trips through the binary format.

DougGregor (changes requested):
Please add a use of this new field in `test/APINotes/Inputs/roundtrip.apinotes` . I suspect it'll expose an issue due to missing a mapping from the 'Types' representation back to YAML.

lib/APINotes/APINotesWriter.cpp
jrose-apple: I see it in the serialization for ObjCContextInfo, [right above](https://github.com/apple/swift-clang/blob/stable/lib/APINotes/APINotesWriter.cpp#L582-L587). Want me to change that one too?

jrose-apple:
Updated in response to feedback.

lib/APINotes/APINotesWriter.cpp
DougGregor: Oh. Huh. I guess we aren't too rigid about this one.

DougGregor (approved):
LGTM, thanks!

lib/APINotes/APINotesWriter.cpp
jrose-apple: It's only because it's a single byte of flags that it's okay. I admit it's not very principled, but writing a single byte as little-endian also seems weird.

PR #41, opened 2016-11-12
[API Notes] Add support for overriding the types of declarations via API notes.

DougGregor:
Introduces the "Type" key for global variables, properties, and
parameters and the "ResultType" key for functions and methods, to
describe the (Objective-)C type of that entity. 

Lazily parse the types provided in these API notes, and apply them to the
AST. The lazy parsing is handled via a callback from Sema into the
Parser, which parses the type in the current context. We perform some
simplistic sanity checking (e.g., the size of the replacement type is
the same as the original type), but otherwise this is a dangerous
power tool that will need to be used carefully.

Addresses rdar://problem/28943642.

### DIFF
--- a/lib/APINotes/APINotesYAMLCompiler.cpp
+++ b/lib/APINotes/APINotesYAMLCompiler.cpp
@@ -173,6 +173,7 @@ namespace {
     unsigned Position;
     Optional<bool> NoEscape = false;
     llvm::Optional<NullabilityKind> Nullability;
+    StringRef Type;
   };
   typedef std::vector<Param> ParamsSeq;
 
@@ -189,6 +190,7 @@ namespace {
       = api_notes::FactoryAsInitKind::Infer;
     bool DesignatedInit = false;
     bool Required = false;
+    StringRef ResultType;
   };
   typedef std::vector<Method> MethodsSeq;
 
@@ -200,6 +202,7 @@ namespace {
     Optional<bool> SwiftPrivate;
     StringRef SwiftName;
     Optional<bool> SwiftImportAsAccessors;
+    StringRef Type;
   };
   typedef std::vector<Property> PropertiesSeq;
 
@@ -224,6 +227,8 @@ namespace {
     AvailabilityItem Availability;
     Optional<bool> SwiftPrivate;
     StringRef SwiftName;
+    StringRef Type;
+    StringRef ResultType;
   };
   typedef std::vector<Function> FunctionsSeq;
 
@@ -233,6 +238,7 @@ namespace {
     AvailabilityItem Availability;
     Optional<bool> SwiftPrivate;
     StringRef SwiftName;
+    StringRef Type;
   };
   typedef std::vector<GlobalVariable> GlobalVariablesSeq;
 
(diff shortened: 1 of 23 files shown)


DougGregor:
(fixes coming in a moment)

lib/APINotes/APINotesYAMLCompiler.cpp
```diff
@@ -385,6 +391,7 @@ namespace llvm {
         io.mapOptional("Nullability",     p.Nullability, 
                                           AbsentNullability);
         io.mapOptional("NoEscape",        p.NoEscape);
+        io.mapOptional("Type",            p.Type, StringRef(""));
```
jrose-apple, line 394: Nitpick: either `""` or `StringRef()` would have been clearer here.

lib/Parse/ParseDecl.cpp
```diff
@@ -6583,3 +6583,68 @@ bool Parser::TryAltiVecTokenOutOfLine(DeclSpec &DS, SourceLocation Loc,
...
+TypeResult Parser::parseTypeFromString(StringRef typeStr, StringRef context,
+                                       SourceLocation includeLoc) {
+  // Consume (unexpanded) tokens up to the end-of-directive.
+  SmallVector<Token, 4> tokens;
+  {
+    // Create a new buffer from which we will parse the type.
+    auto &sourceMgr = PP.getSourceManager();
+    FileID fileID = sourceMgr.createFileID(
+                      llvm::MemoryBuffer::getMemBufferCopy(typeStr, context),
+                      SrcMgr::C_User, 0, 0, includeLoc);
```
jrose-apple, line 6596: This could cause problems with Swift deciding whether to import the type as a user type or a system type. (I'm annoyed that we still have that difference but we do.)

lib/Sema/SemaAPINotes.cpp
```diff
@@ -264,10 +264,62 @@ static void ProcessAPINotes(Sema &S, Decl *D,
...
 static void ProcessAPINotes(Sema &S, Decl *D,
                             const api_notes::VariableInfo &info,
                             VersionedInfoRole role) {
+  // Type override.
+  if (role != VersionedInfoRole::Versioned &&
+      !info.getType().empty() && S.ParseTypeFromStringCallback) {
+    auto parsedType = S.ParseTypeFromStringCallback(info.getType(),
+                                                    "<API Notes>",
+                                                    D->getLocation());
+    if (parsedType.isUsable()) {
```
jrose-apple, line 293: Nitpick: The nesting here makes me think this would be better factored out into a separate function with early returns.

jrose-apple (commented):
Nothing critical but a few small comments.

lib/APINotes/APINotesYAMLCompiler.cpp
DougGregor: Eh, sure.

lib/Parse/ParseDecl.cpp
DougGregor: That's a good point! I'll need to pass the `CharacteristicKind` through.

lib/Sema/SemaAPINotes.cpp
DougGregor: Yeah, fair point. Thanks!

PR #43, opened 2016-11-14
Handle nullability (_Nonnull, _Nullable) for array-typed parameters

jrose-apple:
Since array parameters decay to pointers, `_Nullable` and friends should be available for use there as well. This is especially important for parameters that are typedefs of arrays. The unsugared syntax for this follows the syntax for const-qualified array parameters in C:

```c
void test(int values[_Nullable]);
```

This syntax was previously accepted but the `_Nullable` (and any other attributes) were silently discarded. However, applying `_Nullable` to a typedef was previously rejected and is now accepted.

All of these commits (except the API notes one and the va_list fix-ups) were reviewed on reviews.llvm.org by @DougGregor. More information in the individual commit messages, particularly c899a30.

rdar://problem/25846421

### DIFF
--- a/include/clang/AST/Type.h
+++ b/include/clang/AST/Type.h
@@ -2263,19 +2263,15 @@ public:
 /// Represents a pointer type decayed from an array or function type.
 class DecayedType : public AdjustedType {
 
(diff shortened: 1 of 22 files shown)


jrose-apple:
Passed Swift-side tests in #5777. Merging.

PR #44, opened 2016-11-18
Fix building specific Swift files in swift-clang with MSVC

hughbe:
### Commit 1:
- MSVC complains: specific_attr_begin is not a member of swift::Decl
- The fix is to extract the default parameter and specifiy it for each call of the method

### Commit 2:
MSVC implodes compiling Clang, owing to the definition of llvm::None:
```
namespace llvm {
/// \brief A simple null object to allow implicit construction of Optional<T>
/// and similar types without having to spell out the specialization's name.
enum class NoneType { None };
const NoneType None = None;
}
```

https://connect.microsoft.com/VisualStudio/feedback/details/3111599/

It throws the following strange, obfuscated error:

```
Error: Unhandled exception at 0x00007FFFF9669633 (ntdll.dll) in cl.exe: 0xC0000005: Access violation writing location 0x00000036C7A00E90.

[1/665] Building CXX object tools\clang\lib\APINotes\CMakeFiles\clangAPINotes.dir\APINotesYAMLCompiler.cpp.obj
FAILED: tools/clang/lib/APINotes/CMakeFiles/clangAPINotes.dir/APINotesYAMLCompiler.cpp.obj
C:\PROGRA~2\MICROS~1.0\VC\bin\amd64\cl.exe   /nologo /TP -DCLANG_ENABLE_ARCMT -DCLANG_ENABLE_OBJC_REWRITER -DCLANG_ENABLE_STATIC_ANALYZER -DGTEST_HAS_RTTI=0 -DUNICODE -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_NONSTDC_NO_WARNINGS -D_CRT_SECURE_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_GNU_SOURCE -D_HAS_EXCEPTIONS=0 -D_SCL_SECURE_NO_DEPRECATE -D_SCL_SECURE_NO_WARNINGS -D_UNICODE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -Itools\clang\lib\APINotes -IC:\Users\hbellamy\Documents\GitHub\llvm\tools\clang\lib\APINotes -IC:\Users\hbellamy\Documents\GitHub\llvm\tools\clang\include -Itools\clang\include -Iinclude -IC:\Users\hbellamy\Documents\GitHub\llvm\include /W4 -wd4141 -wd4146 -wd4180 -wd4244 -wd4258 -wd4267 -wd4291 -wd4345 -wd4351 -wd4355 -wd4456 -wd4457 -wd4458 -wd4459 -wd4503 -wd4624 -wd4722 -wd4800 -wd4100 -wd4127 -wd4512 -wd4505 -wd4610 -wd4510 -wd4702 -wd4245 -wd4706 -wd4310 -wd4701 -wd4703 -wd4389 -wd4611 -wd4805 -wd4204 -wd4577 -wd4091 -wd4592 -wd4319 -wd4324 -w14062 -we4238 /Zc:inline /Zc:strictStrings /Oi /Zc:rvalueCast -O2   -UNDEBUG  /EHs-c- /GR- /showIncludes /Fotools\clang\lib\APINotes\CMakeFiles\clangAPINotes.dir\APINotesYAMLCompiler.cpp.obj /Fdtools\clang\lib\APINotes\CMakeFiles\clangAPINotes.dir\ /FS -c C:\Users\hbellamy\Documents\GitHub\llvm\tools\clang\lib\APINotes\APINotesYAMLCompiler.cpp
Internal Compiler Error in C:\PROGRA~2\MICROS~1.0\VC\bin\amd64\cl.exe.  You will be prompted to send an error report to Microsoft later.
```

The fix is to give MSVC some more information. However, it still complains if we use `llvm::None`, so we have to use `llvm::NoneType::None`. I assume the problem is that MSVC gets horribly confused at this line: `const NoneType None = None;`

## Commit 3
- MSVC cannot resolve the operator.
- The fix is to add a static cast to unsigned: this has been done in other places in the file, and was missing in this case, presumably as an oversight

/cc @DougGregor (this time, these files are swift-specific and no upstream version exists in Clang)


lib/APINotes/APINotesYAMLCompiler.cpp
```diff
@@ -164,7 +164,15 @@ namespace {
     AvailabilityItem() : Mode(APIAvailability::Available), Msg("") {}
   };
 
-  static llvm::Optional<NullabilityKind> AbsentNullability = llvm::None;
+  // There is a bug in MSVC that causes an internal compiler error in compilation.
+  // See: https://connect.microsoft.com/VisualStudio/feedback/details/3111599/
+  // This is a workaround that issue until the problem gets fixed.
+#if defined(_MSC_VER)
+  static llvm::Optional<NullabilityKind> AbsentNullability = llvm::NoneType::None;
```
hyp: I think llvm::Optional is initialised to none by default, so shouldn't something like this be sufficient: `llvm::Optional< NullabilityKind > AbsentNullability {};`? This should allow you to get rid of `#if` checks.

hughbe:
I just added another commit that fixes another MSVC error. This is the 3rd commit - however, unlike the others, it modifies code that already exists in Clang. However, the good news is that the Clang folks must've fixed this, as the Clang code: https://github.com/llvm-mirror/clang/blob/8f035b6f8c06db3477406d9da4b88ccdc4fca880/lib/Sema/SemaType.cpp#L3529 has the fix

lib/APINotes/APINotesYAMLCompiler.cpp
hughbe: Thanks @hyp, I should've tagged you, sorry.
I'll take a look. I do recall that MSVC complained about initializing `Optional<T>` with `{}` in a different code base, but maybe this will work.

hughbe: This worked, thanks Alex.

Curious, the following: `llvm::Optional< NullabilityKind > AbsentNullability = {};` _didn't_ work: I presumed you made a typo msising the `=`. Clearly you didn't :D. Do you know why?

hyp: I think the form with the `=` has different semantics. Not sure how they differ, but I think the compiler might treat it as initializer list initialization.

hughbe: That's possible, cheers

lib/Sema/SemaAPINotes.cpp
```diff
@@ -180,7 +179,7 @@ static void ProcessAPINotes(Sema &S, Decl *D,
         return UnavailableAttr::CreateImplicit(S.Context,
                                                CopyString(S.Context,
                                                           info.UnavailableMsg));
-    });
+    }, [](Decl *decl) { return decl->specific_attr_begin<UnavailableAttr>(); });
```
jrose-apple: Really not happy with this duplication. If MSVC really can't handle the default argument, maybe you can make an overload instead?

lib/Sema/SemaType.cpp
```diff
@@ -3522,7 +3522,7 @@ static void recordNullabilitySeen(Sema &S, SourceLocation loc) {
     S.Diag(fileNullability.PointerLoc, diag::warn_nullability_missing_array);
   } else {
     S.Diag(fileNullability.PointerLoc, diag::warn_nullability_missing)
-      << fileNullability.PointerKind;
+      << static_cast<unsigned>(fileNullability.PointerKind);
```
jrose-apple, line 3525: This has been fixed upstream; I just hadn't cherry-picked it. Can you take the real commit instead?

jrose-apple (commented):
Thanks! Some process notes: nothing new should actually be committed to the 'stable' branch. Instead, changes should go to the 'upstream-with-swift' branch and then cherry-picked to the latest release branch if necessary—in this case 'swift-3.1-branch'. This will then get automerged into 'stable'.

As we get further into a release, we'll start locking down more heavily on cherry-picks to the release branch, requiring them to [follow a review template and be approved by a release manager](https://swift.org/blog/swift-3-0-release-process/) (not just a code owner). But for now we're still early enough in Swift 3.1 that a regular review should suffice.

hughbe:
> Some process notes: nothing new should actually be committed to the 'stable' branch.

Sorry, there's not much documentation about the contribution process to these more niche repos. Would you like me to change the target branch? I think that might involve starting a new PR (so maybe the benefits don't outweigh the costs)

hyp:
@hughbe I think you can change the target branch without creating a new PR using the 'edit' button (at least I can).

jrose-apple:
We/I keep meaning to write down the branch structure somewhere convenient. I'll take this as a reminder to do that after (American) Thanksgiving.

hughbe:
@hyp  good in theory,  in practice GitHub just destroyed this PR :D

hughbe:
I'm gonna have to close this PR, and I screwed up again - sorry all for the double notifications. First one was GitHub's fault, second was mine :D

PR #47, opened 2016-11-28
Add -Wapi-notes-message diagnostic group

hyp:
I will be working on ensuring that all warnings in clang have corresponding -W flags (rdar://8104080), and would like to fix this one first as it's not upstreamed and I would like to avoid future merge conflicts.

Cheers,
Alex


hyp:
@DougGregor

jrose-apple:
I'll let Doug be the final judge, but it seems to me that the flag should just be -Wapinotes, to match the -fapinotes-* options. The fact that we just substitute in a string to use as the entire diagnostic seems like an implementation detail.

DougGregor:
I agree that it should just be `-Wapinotes`.

DougGregor:
(with that change, LGTM)

hyp:
I forgot to update the commit message for "-Wapinotes", will recommit.

PR #53, opened 2016-12-13
[APINotes] Preserve attributes from inactive versions.

jrose-apple:
...by wrapping them in another (new) attribute, SwiftVersionedAttr. This can track both attributes that would be added and those that would be removed.

This doesn't support other kinds of changes that can be made via API notes, such as nullability or wholesale type changes, but it's a place to start, and possibly sufficient for our goals.

Part of rdar://problem/28618121.

### DIFF
--- a/lib/Sema/SemaAPINotes.cpp
+++ b/lib/Sema/SemaAPINotes.cpp
@@ -17,6 +17,21 @@
 using namespace clang;
 using clang::api_notes::VersionedInfoRole;
 
+namespace {
+  struct VersionedInfoMetadata {
+    VersionTuple Version;
+    VersionedInfoRole Role;
+
+    /*implicit*/ VersionedInfoMetadata(VersionedInfoRole role) : Role(role) {
+      assert(role != VersionedInfoRole::Versioned &&
+             "explicit version required");
+    }
+
+    /*implicit*/ VersionedInfoMetadata(VersionTuple version)
+        : Version(version), Role(VersionedInfoRole::Versioned) {}
+  };
+} // end anonymous namespace
+
 /// Determine whether this is a multi-level pointer type.
 static bool isMultiLevelPointerType(QualType type) {
   QualType pointee = type->getPointeeType();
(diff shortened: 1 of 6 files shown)


jrose-apple:
Also cherry-pick a fix that accidentally went into 'stable' instead of upstream-with-swift. @hughbe, can you cherry-pick that to swift-3.1-branch too?

jrose-apple:
Factored that out into #54.

DougGregor (approved):
This is simpler than I'd expected, which is fantastic. LGTM!

hughbe:
@jrose-apple I'm fairly sure these changes went into swift-3.1-branch... see: https://github.com/apple/swift-clang/pull/45

Also, the apple/swift-3.1-branch seems up to date with my changes

jrose-apple:
Ah, strange. I guess I checked the wrong thing with git. We did miss upstream-with-swift, though—oops!

lib/Sema/SemaAPINotes.cpp
```diff
@@ -149,8 +171,14 @@ namespace {
...
       if (existing != end) {
-        // Remove the existing attribute.
+        // Remove the existing attribute, and treat it as a superseded
+        // non-versioned attribute.
+        auto *versioned =
+            SwiftVersionedAttr::CreateImplicit(S.Context, clang::VersionTuple(),
+                                               *existing);
+
         D->getAttrs().erase(existing.getCurrent());
+        D->addAttr(versioned);
```
jrose-apple, line 181: @DougGregor: This part is new since last time. Do you see any issues with this version?

jrose-apple:
@fredriss, what do you think about this? This particular PR is going to 'upstream-with-swift', but I'd like to land the corresponding Swift changes soon after. Is it worth cherry-picking to Clang's 'swift-3.1-branch', or should I limit it to the 'stable' branch and let it get picked up in the next Clang rebranch (presumably for Swift 4)?

lib/Sema/SemaAPINotes.cpp
DougGregor: I guess an empty `VersionTuple` is code for "latest version". That seems fine to me, and this looks like the right way to maintain the attribute that was present in the source code without the rest of Clang (or other tools that only care about the API-note-adjusted view) seeing it.

DougGregor (approved):

PR #67, opened 2017-01-19
[APINotes] Remove VersionedInfoRole

jrose-apple:
When API notes were first added, they would defer to the header when something was explicitly specified---for example, if the API notes had one set of nullability annotations from before the header was audited. This behavior didn't make sense for versioned API notes, however, where annotations needed for Swift 3 compatibility were intended to override the unversioned information in both the API notes and the header itself.

Now that the versioning of API notes is well-established, simplify things by having the API notes always "win"; that is, an annotation in API notes will always override what's in the headers.

While here, remove the obsolete `FactoryAsInit` key from the API notes dictionary. This never had a good implementation with versioned API notes anyway.

Getting this into a release branch or 'stable' depends on apple/swift#6901 and apple/swift#6907.

### DIFF
--- a/lib/APINotes/APINotesWriter.cpp
+++ b/lib/APINotes/APINotesWriter.cpp
@@ -766,7 +766,7 @@ namespace {
     }
 
     void emitUnversionedInfo(raw_ostream &out, const ObjCMethodInfo &info) {
-      uint8_t payload = info.FactoryAsInit;
+      uint8_t payload = 0;
       payload = (payload << 1) | info.DesignatedInit;
       payload = (payload << 1) | info.Required;
       endian::Writer<little> writer(out);
--- a/include/clang/APINotes/APINotesReader.h
+++ b/include/clang/APINotes/APINotesReader.h
@@ -25,20 +25,6 @@
 namespace clang {
 namespace api_notes {
 
-/// Describes the role of a specific bit of versioned information.
-enum class VersionedInfoRole : unsigned {
-  /// Augment the AST, but do not override information explicitly specified
-  /// in the source code.
-  AugmentSource,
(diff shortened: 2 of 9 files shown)


lib/APINotes/APINotesWriter.cpp
```diff
@@ -766,7 +766,7 @@ namespace {
     }
 
     void emitUnversionedInfo(raw_ostream &out, const ObjCMethodInfo &info) {
-      uint8_t payload = info.FactoryAsInit;
+      uint8_t payload = 0;
```
DougGregor, line 769: I see that you kept the API notes binary format the same. Thanks!

lib/Sema/SemaDecl.cpp
```diff
@@ -13537,7 +13537,6 @@ Decl *Sema::ActOnTag(Scope *S, unsigned TagSpec, TagUseKind TUK,
 
   if (Attr)
     ProcessDeclAttributeList(S, New, Attr);
-  ProcessAPINotes(New);
```
DougGregor, line 13540: Why did this go away?

DougGregor (changes requested):
Why is it correct to remove the ProcessAPINotes call from ActOnTag? If it is correct, we should separate it out.

lib/Sema/SemaDecl.cpp
jrose-apple: Whoops! That was supposed to be a separate patch (with a good explanation and a test case). Thanks for catching this!

jrose-apple:
Fixed!

DougGregor (approved):
Thanks!

PR #45, opened 2016-11-18
Fix building specific Swift files in swift-clang with MSVC

hughbe:
Previous PR #44 was nuked by GitHub

### Commit 1:
- MSVC complains: specific_attr_begin is not a member of swift::Decl
- The fix is to extract the default parameter and specifiy it for each call of the method

### Commit 2:
MSVC implodes compiling Clang, owing to the definition of llvm::None:
```
namespace llvm {
/// \brief A simple null object to allow implicit construction of Optional<T>
/// and similar types without having to spell out the specialization's name.
enum class NoneType { None };
const NoneType None = None;
}
```

https://connect.microsoft.com/VisualStudio/feedback/details/3111599/

It throws the following strange, obfuscated error:

```
Error: Unhandled exception at 0x00007FFFF9669633 (ntdll.dll) in cl.exe: 0xC0000005: Access violation writing location 0x00000036C7A00E90.

[1/665] Building CXX object tools\clang\lib\APINotes\CMakeFiles\clangAPINotes.dir\APINotesYAMLCompiler.cpp.obj
FAILED: tools/clang/lib/APINotes/CMakeFiles/clangAPINotes.dir/APINotesYAMLCompiler.cpp.obj
[...]

PR #68, opened 2017-01-19
Fix compiler crash building APINotesYAMLCompiler in swift-clang with MSVC

hughbe:
This seems to have been reset in stable/swift-4.0-branch after the update to LLVM 4.0

This is a port of #45 (the other stuff seems to still be there with MSVC).

This fixes an MSVC crash,

@jrose-apple, please let me know what other branch this should go to. Do I submit another duplicate PR for stable or upstream-with-swift?

### DIFF
--- a/lib/APINotes/APINotesYAMLCompiler.cpp
+++ b/lib/APINotes/APINotesYAMLCompiler.cpp
@@ -164,7 +164,7 @@ namespace {
     AvailabilityItem() : Mode(APIAvailability::Available), Msg("") {}
   };
 
-  static llvm::Optional<NullabilityKind> AbsentNullability = llvm::None;
+  static llvm::Optional<NullabilityKind> AbsentNullability {};
   static llvm::Optional<NullabilityKind> DefaultNullability =
     NullabilityKind::NonNull;
   typedef std::vector<clang::NullabilityKind> NullabilitySeq;


jrose-apple:
The proper process is to put all LLVM/Clang changes into upstream-with-swift first and then cherry-pick them to the necessary release branches.

Rereading the original commit, I think probably the LLVM definition of `None` is wrong; it should be `const NoneType None = NoneType::None;`. Does that help anything or no?

hughbe:
Umm, that is the definition of `None` in LLVM:
> https://github.com/llvm-mirror/llvm/blob/master/include/llvm/ADT/None.h#L24

jrose-apple:
Yes, I know. :-) I'm wondering if it's wrong, and if so if fixing it makes the error go away.

hughbe:
Haha, maybe I misunderstood your comment - you said "it should be ....`, but it already is!

hughbe:
I see now - I'm looking upstream rather than in swift-llvm! https://github.com/apple/swift-llvm/blob/swift-4.0-branch/include/llvm/ADT/None.h#L23

jrose-apple:
Yeah, back in #45 it used a funny self-referential thing. I'm not sure why ours is different.

hughbe:
I was under the impression everything was reset to upstream in the swift-4.0-branch. If this hasn't been updated, what else hasn't?

jrose-apple:
Either this just came in recently, or it's something we changed in upstream-with-swift. Let's see…looks like it is very recent. https://github.com/llvm-mirror/llvm/commit/cf890beb7ef1148556a45d4a61278a320ef53e16

hughbe:
Okay, should I cherry-pick that commit to upstream-with-swift as well as swift-4.0-branch

jrose-apple:
upstream-with-swift really is an "upstream"; it gets automerged frequently. This release with swift-4.0-branch is a little special in that we're also automerging from LLVM's release branch (coincidentally also version 4, I think). So if this does fix the problem, then ideally someone would cherry-pick it upstream.

(I think there'll be a Swift blog post about this soon, but there was at least a [message on swift-dev](https://lists.swift.org/pipermail/swift-dev/Week-of-Mon-20170116/003857.html).)

hughbe:
Awesome. All the upstream/stable/swift-x.y-branch stuff can get a little confusing :D. I assume we close this until upstream-with-swift, followed by stable, followed by swift-x.y-branch (not necessarily in that order) auto updates?

jrose-apple:
Last piece of the puzzle: "stable" is basically an alias for "the current release branch" to make cherry-picking easy. If we ever have a period where something goes into the *next* release but not *this* one, "stable" is where it would go, but so far that hasn't happened since the LLVM/Clang branches are prepared well before the Swift branches.

hughbe:
Thanks for filling me in on this, appreciate it. Wonderfully complicated stuff...

PR #69, opened 2017-01-19
[APINotes] Don't apply API notes to non-definitions of structs.

jrose-apple:
There are two reasons for this:

- It doesn't really seem appropriate for forward-declared structs to pick up attributes.

- In a proper definition `struct X { ... }`, the first `struct X` counts as a declaration, which means the API note attributes can get added ahead of time. Later on, when we try to add them *again,* they'll replace the attribute added at the first point rather than whatever the user wrote in source after the close brace. This also applies if there are any forward declarations before the definition.

There is a downside: API notes now can't affect tag types that are *never* defined (for example, structs that are forward-declared for use in pointers). Any annotations one might want to apply to such a type must be written in source. We don't seem to have any such types in the API notes today, though (tested using the swift-api-digester).

rdar://problem/29981732

### DIFF
--- a/lib/Sema/SemaDecl.cpp
+++ b/lib/Sema/SemaDecl.cpp
@@ -13673,7 +13673,6 @@ CreateNewDecl:
 
   if (Attr)
     ProcessDeclAttributeList(S, New, Attr);
-  ProcessAPINotes(New);
 
   // If this has an identifier, add it to the scope stack.
   if (TUK == TUK_Friend) {
--- a/test/APINotes/Inputs/Frameworks/SimpleKit.framework/Headers/SimpleKit.apinotes
+++ b/test/APINotes/Inputs/Frameworks/SimpleKit.framework/Headers/SimpleKit.apinotes
@@ -1,0 +1,6 @@
+Name: SimpleKit
+Tags:
+- Name: RenamedAgainInAPINotesA
+  SwiftName: SuccessfullyRenamedA
+- Name: RenamedAgainInAPINotesB
+  SwiftName: SuccessfullyRenamedB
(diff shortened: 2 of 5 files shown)


jrose-apple:
Ping. I still think this is the right thing to do, and it affects a spelling someone might try to use with NS_ENUM.

DougGregor (commented):
The second issue you mention makes me wonder if the real problem here is in attribute merging amongst different declarations of the same entity. There, we don't distinguish between API-notes--generated and written-in-source attributes, so we won't get the same overriding behavior we would expect.

I'm on the fence about this one, because we have the same issue with Objective-C `@class`. There, I vaguely remember adding some annotations for forward declarations of Objective-C classes where we had layering violations in the SDK---one example is the `SwiftBridge` API notes for `NSString` in the `ObjectiveC` module. I think the difference between that case and your experiment with structs is that Swift's Clang importer tolerates imports of forward-declared classes better than forward-declared structs or enums.

jrose-apple:
Both of them should be resolving the same way now: preferring a definition over a declaration, but able to handle forward declarations as long as the defining module is loaded before we start importing any decls. We could certainly distinguish between API-note-created attributes and attributes that are implicit for some other reason, but I'm not sure it's worth it.

DougGregor:
Okay, that's a fine way to handle it, and I'm glad to avoid having to touch the attribute-merging code. Thanks for the clarification

jrose-apple:
Honestly, I'd like to remove the API notes for forward-declarations in the ObjectiveC module as well. Those belong in the Foundation module.

jrose-apple:
I went and did that in apple/swift#8328. Turns out we weren't using them. Then I checked if API notes on a forward declaration of a class can affect how the class is imported, and it looks like they don't. So I think I'm good to merge this!

DougGregor:
Okay, great!

PR #70, opened 2017-01-21
[APINotes] Add a 'SwiftImportAsNonGeneric' entry for ObjC classes

jrose-apple:
If true, Swift will ignore any 'lightweight generics' parameters when importing the class, substituting any uses of them with their bounds. At least, it will once this lands and the Swift side gets implemented.

Finishes rdar://problem/28455962.

### DIFF
--- a/lib/APINotes/APINotesWriter.cpp
+++ b/lib/APINotes/APINotesWriter.cpp
@@ -584,8 +584,12 @@ namespace {
       emitCommonTypeInfo(out, info);
 
       uint8_t payload = 0;
+      if (auto swiftImportAsNonGeneric = info.getSwiftImportAsNonGeneric()) {
+        payload |= (0x01 << 1) | swiftImportAsNonGeneric.getValue();
+      }
+      payload <<= 3;
       if (auto nullable = info.getDefaultNullability()) {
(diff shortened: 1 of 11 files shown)


jrose-apple:
Mostly needs a rubber stamp from Doug.

lib/APINotes/APINotesWriter.cpp
```diff
@@ -584,8 +584,12 @@ namespace {
       emitCommonTypeInfo(out, info);
 
       uint8_t payload = 0;
+      if (auto swiftImportAsNonGeneric = info.getSwiftImportAsNonGeneric()) {
+        payload |= (0x01 << 1) | swiftImportAsNonGeneric.getValue();
+      }
+      payload <<= 3;
```
DougGregor, line 590: Why are we shifting by 3 instead of 2? There are two bits of data here...

DougGregor (approved):

lib/APINotes/APINotesWriter.cpp
jrose-apple: This is making room for the *next* chunk of data, not moving out the *previous* data.

DougGregor: Aaah, right. Guess I should just rubber-stamp. Thanks!

PR #73, opened 2017-03-02
Promote ConstantInitBuilder to be a public CodeGen API

rjmccall:


rjmccall:
Ah, now I remember how to do this.

PR #74, opened 2017-03-16
DO NOT MERGE YET: [ASTReader] Add a DeserializationListener callback for attaching to a new ASTReader

graydon:
This is the clang-part of apple/swift#8143. Needs to either land here or upstream before that swift change will work.


jrose-apple:
I'd definitely want some Clang people to weigh in on this, and the easiest way to do that might be to propose it upstream at reviews.llvm.org. Ben Langmuir, Bruno Lopes, and Richard Smith would be good people to loop in, with a cc to cfe-commits.

(Reminder that if it does turn out to be the way to go, `stable` is not the branch to merge it to. It would need to go in llvm.org or upstream-with-swift and then cherry-picked to swift-4.0-branch.)

graydon:
This has been superseded by https://reviews.llvm.org/D31378, closing

PR #81, opened 2017-05-05
[APINotes] Always provide an unversioned SwiftName of some kind.

jrose-apple:
If the active API notes set a Swift name for a declaration, but no other versions do, it's assumed that that's the name in every version, because the AST is indistinguishable from just applying the attribute in the header. Fix this by explicitly adding an inactive, unversioned removal attribute for SwiftName in otherwise ambiguous cases.

(This really relates to *all* annotations in the "SwiftVersions" section, and particularly those that are usually not written in headers, but it's particularly bad for SwiftName because the Swift compiler actually looks at the inactive versions for that one. At some point it may be worth trying to generalize this, though.)

rdar://problem/31826517

### DIFF
--- a/lib/Sema/SemaAPINotes.cpp
+++ b/lib/Sema/SemaAPINotes.cpp
@@ -649,6 +649,49 @@ static void ProcessAPINotes(Sema &S, ObjCInterfaceDecl *D,
                   metadata);
 }
 
+/// If we're applying API notes with an active, non-default version, and the
+/// versioned API notes have a SwiftName but the declaration normally wouldn't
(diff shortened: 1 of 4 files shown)


jrose-apple:
I actually still need to test the Swift side of this but I'm pretty confident I got it right. This is the same attribute used when NS_SWIFT_NAME is written in the source.

jrose-apple:
Whoops, no, I didn't get it right. I forgot to actually test the case where NS_SWIFT_NAME is written in the source.

DougGregor (approved):
LGTM, thank you!

PR #82, opened 2017-05-09
[4.0] [APINotes] Always provide an unversioned SwiftName of some kind.

jrose-apple:
- **Explanation:** We use API notes to record whether a particular C/ObjC API has changed its name for Swift 4. However, if it had a custom name in Swift 3 and then went back to the default in Swift 4, the Swift 4 name would not be visible in Swift 3 mode. (This is a correctness issue for types, which we always import under their Swift 4 name.) This patch fixes that by explicitly recording the absence of a custom name.
- **Scope:** Fairly narrow; the situation described doesn't seem common. The Swift-side change affects the order things are printed in, but only the order.
- **Issue:** rdar://problem/31826517
- **Reviewed by:** @DougGregor (Clang-side change), @nkcsgexi (Swift helper change)
- **Risk:** Low. Does not affect non-API-notes Clang at all, and uses existing code paths on the Swift side. One minor change to the way declarations are printed in Swift is needed for consistent behavior across platforms.
- **Testing:** Added compiler regression tests, verified that the original test case now behaves as expected.

### DIFF
--- a/lib/Sema/SemaAPINotes.cpp
+++ b/lib/Sema/SemaAPINotes.cpp
@@ -649,6 +649,49 @@ static void ProcessAPINotes(Sema &S, ObjCInterfaceDecl *D,
                   metadata);
 }
 
+/// If we're applying API notes with an active, non-default version, and the
+/// versioned API notes have a SwiftName but the declaration normally wouldn't
+/// have one, add a removal attribute to make it clear that the new SwiftName
+/// attribute only applies to the active version of \p D, not to all versions.
+///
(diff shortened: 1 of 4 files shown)


jrose-apple:
@swift-ci Please test

jrose-apple:
Hm, I may have *still* missed something.

jrose-apple:
apple/swift#9414
@swift-ci Please test macOS

jrose-apple:
Note: this passed full testing in apple/swift#9453.

jrose-apple:
Aack, wrong branch.

PR #83, opened 2017-05-11
[System z]Update default CPU for s390x to support conversion between unsigned i…

vivkong:
…ntegers and floating point numbers

On System z, instructions for conversion between unsigned integers and floating point numbers was added in z196.  We want to have this support by the default on Swift.  

The missing support was exposed by the test swift/validation-test/stdlib/FixedPointConversion.swift.gyb.  With this change, the test case passes on s390x.

### DIFF
--- a/lib/Basic/Targets.cpp
+++ b/lib/Basic/Targets.cpp
@@ -7046,7 +7046,7 @@ class SystemZTargetInfo : public TargetInfo {
 
 public:
   SystemZTargetInfo(const llvm::Triple &Triple, const TargetOptions &)
-      : TargetInfo(Triple), CPU("z10"), HasTransactionalExecution(false),
+      : TargetInfo(Triple), CPU("z196"), HasTransactionalExecution(false),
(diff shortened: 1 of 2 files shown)


hyp:
Should this be fixed upstream?

vivkong:
We only want to change the default in Swift.  Can keep the default unchanged in LLVM/Clang

hyp:
Why is that? I think in general we prefer to avoid such specific changes in swift-clang.

vivkong:
If it's possible, we would also like to include this change in swift-3.1-branch.

DougGregor:
This change should either be made upstream or requires a different solution. Swift's version of LLVM/Clang is only supposed to have additive, Swift-only features---not changes to how Clang works.

vivkong:
Thanks for the review.  I understand why this shouldn't go here.  I've created another PR https://github.com/apple/swift/pull/9645 - hope that is a proper place to fix this.

PR #86, opened 2017-05-20
[index] Fix forward declarations interfering with USR generation of external source symbols

akyrtzi:
Patch by Nathan Hawes.
https://reviews.llvm.org/D33346

From commit r303484


akyrtzi:
Explanation: If there is a @class forward reference before the @interface annotated with external_source_symbol attribute, clang misses the attribute because clang considers the first occurrence of @class as the 'canonical' decl, but that one doesn't have the attribute.
Scope: Only affects the indexing code path in clang and libclang
Radar: rdar://32271627
Risk: Low
Testing: Added regression test

PR #88, opened 2017-06-20
[refactor] "Generate Missing Function Definitions" should generate st…

hyp:
…ubs in @implementation for a declaration in a class extension

rdar://32869347


vedantk (approved):
LGTM

PR #89, opened 2017-06-20
[refactor] Fix a crash that happens when generating a "Add Missing Switch Cases" fixit while parsing an incomplete switch

hyp:
rdar://32854328


vedantk (approved):
LGTM.

PR #90, opened 2017-06-21
[refactor] "Fill in abstract methods" should not place method groups after an  implicit method that overrides a method in the abstract class

hyp:
This commit fixes an issue where the refactoring action "Fill in abstract methods"
inserted method declarations outside of the class body because Clang decided to
place them after the implicit destructor.

rdar://32857076


vedantk (approved):
LGTM

PR #93, opened 2017-06-23
[Lex] Expose a ModuleMacro's IdentifierInfo.

jrose-apple:
Support for apple/swift#10519.

### DIFF
--- a/include/clang/Lex/MacroInfo.h
+++ b/include/clang/Lex/MacroInfo.h
@@ -510,6 +510,9 @@ public:
     ID.AddPointer(II);
   }
 
+  /// Get the name of the macro.
+  IdentifierInfo *getName() const { return II; }
+
   /// Get the ID of the module that exports this macro.
   Module *getOwningModule() const { return OwningModule; }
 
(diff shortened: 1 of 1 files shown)


include/clang/Lex/MacroInfo.h
```diff
@@ -515,6 +510,9 @@ class ModuleMacro : public llvm::FoldingSetNode {
     ID.AddPointer(II);
   }
 
+  /// Get the name of the macro.
+  IdentifierInfo *getName() const { return II; }
```
jrose-apple, line 514: @zygoloid, I'll bring this to Clang proper if it makes sense, but *does* it make sense? It seems like if we have this information it should be reasonable to ask for it, but if not I'll figure out another way to refer to "this particular macro definition by this name" in our lookup table. (Which doesn't *have* to keep pointer-sized entries, but it'd be nice if it could.)

bob-wilson:
The Swift WIP PR was merged, so I'm going to merge this one as well so that we can make progress on other issues. At least on the Swift side, there is more work to be done on this, though.

include/clang/Lex/MacroInfo.h
zygoloid: Yes, it seems fine to expose this; the name is part of the module macro abstraction and not just an implementation detail. I'm a little surprised we don't provide such an accessor already :)

PR #95, opened 2017-06-27
[index/build] Upstream Clang index-while-building

bitjammer:
This patch upstreams previously AppleInternal index-while-building
changes released in the Xcode 9 timeframe.


akyrtzi (approved):
LGTM.

PR #96, opened 2017-06-29
Rdar 30851899 bridging pch mergemodule mismatch swift 4.0 branch

graydon:
**Explanation**: Fix a bridging PCH crash caused by a clang bug

**Scope**: Causes failure-to-compile (appears as a deserialization error) unless user disables bridging PCH. Triggered by legal-but-clumsy misconfiguration in workspace, seemingly as-performed by cocoapods in unknown number of cases. Might be quite widespread.

**Radar**: 30851899

**Risk**: relatively low, it's not even clear to me that a user can detect it in C/ObjC, and it only makes an under-defined corner case of the module system a bit more-deterministic; it's just a corner-case that swift happens to hit when importing definitions from a nest of header symlinks, as produced by cocoapods.

**Testing**: Swift has a pending test for it https://github.com/apple/swift/pull/10658 ; at the clang level, it can't easily be checked in ObjC, but the misbehaviour can trigget an assertion in clang, which there's an accompanying test for here.

### DIFF
--- a/include/clang/AST/Redeclarable.h
+++ b/include/clang/AST/Redeclarable.h
@@ -21,6 +21,60 @@
 namespace clang {
 class ASTContext;
 
+// Some notes on redeclarables:
+//
+//  - Every redeclarable is on a circular linked list.
+//
+//  - Every decl has a pointer to the first element of the chain _and_ a
+//    DeclLink that may point to one of 3 possible states:
+//      - the "previous" (temporal) element in the chain
+//      - the "latest" (temporal) element in the chain
+//      - the an "uninitialized-latest" value (when newly-constructed)
+//
+//  - The first element is also often called the canonical element. Every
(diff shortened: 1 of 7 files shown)


graydon:
@swift-ci please test

include/clang/AST/Redeclarable.h
```diff
@@ -21,6 +21,60 @@
...
+//      |            |       |              |       |              |
+//      +------------+       +--------------+       +--------------+
+//      | data       | defn  | data         |  defn | data         |
+//      |            | ----> |              | <---- |              |
+//      +------------+       +--------------+       +--------------+
+//        |                     |     ^                  ^
+//        |                     |defn |                  |
+//        | link                +-----+                  |
+//        +-->-------------------------------------------+
+
```
bcardosolopes, line 77: This is very nice!

bcardosolopes (approved):

PR #98, opened 2017-06-30
Look through declarations that start in a macro but end up in a source file when searching for a declaration at some location

hyp:
Otherwise, rename can't be initiated inside of an NS_ENUM

rdar://32934347


vedantk (approved):
I like the test. LGTM.

hyp:
Committed To https://github.com/apple/swift-clang
   8a4f612879..6ea0e95dbd  upstream-with-swift -> upstream-with-swift

PR #99, opened 2017-07-06
[swift-4.0-branch] [objc] Don't emit memset when result is ignored for struct-returning method calls

kubamracek:
This is a cherry-pick/back-port of a part of r306838, "[objc] Don't require null-check and don't emit memset when result is ignored for struct-returning method calls". Fixes an issue with the emission of lifetime markers for struct-returning Obj-C msgSend calls.


rjmccall (approved):
I don't have any technical objection to this patch.

ahatanaka (approved):

PR #101, opened 2017-07-10
[refactor][index] Handle ObjC method names with empty selector pieces

hyp:
1st commit: will go to upstream.
The other two: on GitHub.com/apple/swift-clang.

### DIFF
--- a/lib/Tooling/Refactor/RenameIndexedFile.cpp
+++ b/lib/Tooling/Refactor/RenameIndexedFile.cpp
@@ -81,14 +81,6 @@ static MatchKind checkOccurrence(const IndexedOccurrence &Occurrence,
   if (BeginLoc.isInvalid())
     return MatchKind::None;
   StringRef SymbolNameStart = Symbol.Name[0];
-  SourceLocation EndLoc = BeginLoc.getLocWithOffset(std::max(
-      SymbolNameStart.size(),
-      size_t(1))); // Take empty Objective-C selector pieces into account.
-  if (!SM.isBeforeInTranslationUnit(BeginLoc, EndLoc)) {
-    // Ignore any invalid source ranges. This can occur if the indexed
-    // location is invalid.
-    return MatchKind::None;
-  }
   // Extract the token at the location.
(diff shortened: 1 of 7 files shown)


lib/Tooling/Refactor/RenameIndexedFile.cpp
```diff
@@ -81,7 +81,9 @@ static MatchKind checkOccurrence(const IndexedOccurrence &Occurrence,
   if (BeginLoc.isInvalid())
     return MatchKind::None;
   StringRef SymbolNameStart = Symbol.Name[0];
-  SourceLocation EndLoc = BeginLoc.getLocWithOffset(SymbolNameStart.size());
+  SourceLocation EndLoc = BeginLoc.getLocWithOffset(std::max(
+      SymbolNameStart.size(),
+      size_t(1))); // Take empty Objective-C selector pieces into account.
   if (!SM.isBeforeInTranslationUnit(BeginLoc, EndLoc)) {
```
benlangmuir: Is `isBeforeInTranslationUnit` still correct after the above change?

hyp: Yeah, the location just after ':' will be after the location of ':'.

benlangmuir: I guess what I was asking is: won't it always be after, since the offset >= 1 now instead of >= 0?

hyp: Ah. Right. It looks like we don't need this check at all now. The initial implementation used `Lexer::getSourceText(
      CharSourceRange::getCharRange(BeginLoc, EndLoc)` so we needed to verify the range, but now we just lex the token so this is redundant.

hyp:
Updated.

benlangmuir (approved):
LGTM

hyp:
First commit - r307901

hyp:
To https://github.com/apple/swift-clang
   2bafda8fb0..13bfdc79c2  upstream-with-swift -> upstream-with-swift

PR #102, opened 2017-07-12
[refactor] "Generate Missing Function Definitions" should insert

hyp:
category methods into class implementation declarations even when
the category is not available in the TU in which the class is implemented

The previous PR at https://github.com/apple/swift-clang/pull/94 didn't take into account that the category might not be declared in the TU which has the classes implementation. The non-defined methods will resolve to NULL in that case and the operation won't actually create the method stubs. This patch ensures that method stubs are pre-generated in the initiation TU and that the result of the not-defined query is propagated to the continuation even when the method resolves to NULL (In a hacky way, but it will be improved in the new engine at llvm.org).

rdar://32875896

### DIFF
--- a/include/clang/Tooling/Refactor/IndexerQuery.h
+++ b/include/clang/Tooling/Refactor/IndexerQuery.h
@@ -181,6 +181,33 @@ public:
 
 } // end namespace detail
 
+enum class QueryBoolResult {
+  Unknown,
+  Yes,
+  No,
+};
+
+// FIXME: Either PersistentDeclRef or Decl *
+template <typename T> struct Indexed {
+  T Decl;
+  // FIXME: Generalize better in the new refactoring engine.
+  QueryBoolResult BoolResult = QueryBoolResult::Unknown;
+
+  Indexed(T Decl, QueryBoolResult BoolResult = QueryBoolResult::Unknown)
+      : Decl(Decl), BoolResult(BoolResult) {}
+
+  Indexed(Indexed<T> &&Other) = default;
+  Indexed &operator=(Indexed<T> &&Other) = default;
+  Indexed(const Indexed<T> &Other) = default;
+  Indexed &operator=(const Indexed<T> &Other) = default;
+
+  /// True iff the declaration is not defined in the entire project.
+  bool isNotDefined() const {
+    // FIXME: This is hack. Need a better system in the new engine.
+    return BoolResult == QueryBoolResult::Yes;
+  }
+};
+
 /// Transforms one set of declarations into another using some predicate.
 class DeclarationsQuery : public IndexerQuery {
   static const char *BaseUIDString;
@@ -189,7 +216,7 @@ class DeclarationsQuery : public IndexerQuery {
   std::unique_ptr<detail::DeclPredicateNode> Predicate;
 
 protected:
-  std::vector<PersistentDeclRef<Decl>> Output;
+  std::vector<Indexed<PersistentDeclRef<Decl>>> Output;
 
 public:
   DeclarationsQuery(std::vector<const Decl *> Input,
@@ -203,7 +230,7 @@ public:
 
   void invalidateTUSpecificState() override { Input.clear(); }
 
-  void setOutput(std::vector<PersistentDeclRef<Decl>> Output) {
+  void setOutput(std::vector<Indexed<PersistentDeclRef<Decl>>> Output) {
     this->Output = Output;
   }
 
@@ -238,10 +265,11 @@ public:
       : DeclarationsQuery(std::vector<const Decl *>(Input.begin(), Input.end()),
                           std::move(Predicate)) {}
 
-  std::vector<PersistentDeclRef<T>> getOutput() const {
-    std::vector<PersistentDeclRef<T>> Results;
+  std::vector<Indexed<PersistentDeclRef<T>>> getOutput() const {
+    std::vector<Indexed<PersistentDeclRef<T>>> Results;
     for (const auto &Ref : DeclarationsQuery::Output)
-      Results.push_back(PersistentDeclRef<T>(Ref.USR));
+      Results.push_back(Indexed<PersistentDeclRef<T>>(
+          PersistentDeclRef<T>(Ref.Decl.USR), Ref.BoolResult));
     return Results;
   }
 };
--- a/include/clang/Tooling/Refactor/IndexerQuery.h
+++ b/include/clang/Tooling/Refactor/IndexerQuery.h
@@ -187,14 +187,14 @@ enum class QueryBoolResult {
   No,
 };
 
-// FIXME: Either PersistentDeclRef or Decl *
+// FIXME: Check that 'T' is either a PersistentDeclRef<> or a Decl *.
 template <typename T> struct Indexed {
   T Decl;
   // FIXME: Generalize better in the new refactoring engine.
-  QueryBoolResult BoolResult = QueryBoolResult::Unknown;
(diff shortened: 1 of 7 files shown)


lib/Tooling/Refactor/ImplementDeclaredMethods.cpp
```diff
@@ -408,21 +426,15 @@ ImplementDeclaredObjCMethodsOperation::runInImplementationAST(
...
-  PP.PolishForDeclaration = true;
-  PP.SuppressStrongLifetime = true;
-  PP.SuppressLifetimeQualifiers = true;
-  PP.SuppressUnwrittenScope = true;
-  for (const ObjCMethodDecl *MD : SelectedMethods) {
+  assert(MethodDeclarations.size() >= SelectedMethods.size() &&
+         "less declarations than selected methods?");
+  for (const auto &I : llvm::enumerate(SelectedMethods)) {
+    indexer::Indexed<const ObjCMethodDecl *> Decl = I.value();
     // Skip methods that are already defined.
```
vedantk: Since Decl.BoolResult could be Unknown, are we also skipping definitions that 'may' already be defined?

lib/Tooling/Refactor/ImplementDeclaredMethods.cpp
```diff
@@ -408,21 +426,15 @@ ImplementDeclaredObjCMethodsOperation::runInImplementationAST(
...
   llvm::raw_string_ostream OS(MethodString);
 
-  PrintingPolicy PP = Context.getPrintingPolicy();
-  PP.PolishForDeclaration = true;
-  PP.SuppressStrongLifetime = true;
-  PP.SuppressLifetimeQualifiers = true;
-  PP.SuppressUnwrittenScope = true;
-  for (const ObjCMethodDecl *MD : SelectedMethods) {
+  assert(MethodDeclarations.size() >= SelectedMethods.size() &&
+         "less declarations than selected methods?");
```
vedantk: nit, fewer

include/clang/Tooling/Refactor/IndexerQuery.h
```diff
@@ -181,6 +181,33 @@ class DeclPredicateNotPredicate : public DeclPredicateNode {
...
+  Unknown,
+  Yes,
+  No,
+};
+
+// FIXME: Either PersistentDeclRef or Decl *
+template <typename T> struct Indexed {
+  T Decl;
+  // FIXME: Generalize better in the new refactoring engine.
+  QueryBoolResult BoolResult = QueryBoolResult::Unknown;
```
vedantk: I think using a more descriptive name here like 'IsDefined' would make the code more readable.

include/clang/Tooling/Refactor/IndexerQuery.h
```diff
@@ -181,6 +181,33 @@ class DeclPredicateNotPredicate : public DeclPredicateNode {
 
 } // end namespace detail
 
+enum class QueryBoolResult {
+  Unknown,
+  Yes,
+  No,
+};
+
+// FIXME: Either PersistentDeclRef or Decl *
```
vedantk: This could use a slightly more descriptive comment.

include/clang/Tooling/Refactor/IndexerQuery.h
```diff
@@ -181,6 +181,33 @@ class DeclPredicateNotPredicate : public DeclPredicateNode {
...
+template <typename T> struct Indexed {
+  T Decl;
+  // FIXME: Generalize better in the new refactoring engine.
+  QueryBoolResult BoolResult = QueryBoolResult::Unknown;
+
+  Indexed(T Decl, QueryBoolResult BoolResult = QueryBoolResult::Unknown)
```
vedantk: It doesn't look like we take advantage of the fact that BoolResult has a default value. Maybe it should be dropped?

tools/libclang/CRefactor.cpp
```diff
@@ -1884,13 +1884,18 @@ clang_IndexerQuery_consumeIntResult(CXIndexerQuery Query, unsigned CursorIndex,
...
   bool IsNot = false;
   if (isa<indexer::detail::DeclPredicateNotPredicate>(DQ->getPredicateNode()))
     IsNot = true;
-  Wrapper->DeclResults[CursorIndex] = PersistentDeclRef<Decl>::create(
-      (IsNot ? !Value : !!Value) ? DQ->getInputs()[CursorIndex] : nullptr);
+  bool Result = IsNot ? !Value : !!Value;
+  Wrapper->DeclResults[CursorIndex] = indexer::Indexed<PersistentDeclRef<Decl>>(
+      PersistentDeclRef<Decl>::create(Result ? DQ->getInputs()[CursorIndex]
+                                             : nullptr),
+      Result ? indexer::QueryBoolResult::Yes : indexer::QueryBoolResult::No);
```
vedantk, line 1897: Just to double check: if you construct an Indexed(DQ->getInputs()[CursorIndex], QueryBoolResult::Yes), are you saying that the declaration is not defined in the entire project?

lib/Tooling/Refactor/RefactoringContinuations.h
```diff
@@ -154,13 +165,46 @@ class PersistentToASTSpecificStateConverter {
     return Results;
   }
 
+  template <typename T>
+  bool addConvertible(
```
vedantk, line 169: Do we ever want to persist PersistentDeclRefs not wrapped in Indexed structs?

lib/Tooling/Refactor/RefactoringContinuations.h
```diff
@@ -154,13 +165,46 @@ class PersistentToASTSpecificStateConverter {
...
+  }
+
   bool addConvertible(const PersistentFileID &) {
     // Do nothing since FileIDs are converted one-by-one.
     return true;
   }
 
   FileID convert(const PersistentFileID &Ref);
 
+  bool addConvertible(const std::vector<std::string> &) { return true; }
```
vedantk, line 202: This is for the method declarations vector, right?

lib/Tooling/Refactor/ImplementDeclaredMethods.cpp
hyp: Yes, but in practice this won't normally happen. Because of how this is implemented, the indexer will either find out whether all the queried methods are defined, or the continuation will receive an empty vector of methods.

include/clang/Tooling/Refactor/IndexerQuery.h
hyp: You're right. I'll update the code to use the default value when constructing this structure.

tools/libclang/CRefactor.cpp
hyp: Yep, it's confusing because we're querying if `! isDefined`, and `YES` refers to `! isDefined`, and not `isDefined`.

lib/Tooling/Refactor/RefactoringContinuations.h
hyp: No, since they're only produced by queries and aren't state. Therefore we just need to convert them back to `Decl *`.

lib/Tooling/Refactor/RefactoringContinuations.h
hyp: Yep.

vedantk (approved):
Thanks, lgtm.

hyp:
Thanks, To https://github.com/apple/swift-clang
   d2e5cf7dd6..c790745221  upstream-with-swift -> upstream-with-swift

PR #103, opened 2017-07-12
[index] Don't add relation to a NamedDecl with no name 

benlangmuir:
Fixes a crash in index record hashing.

rdar://problem/32474406

*Note*: the first commit will go to clang upstream, the second (test-only) commit will go to swift-clang because it requires the indexing-while-building code.


benlangmuir:
@swift-ci please smoke test

akyrtzi:
LGTM!

PR #105, opened 2017-07-13
The "does not conform to protocol" -Wprotocol refactoring warning should take method availability into account

hyp:
This also ensures that the "Add missing requirements"
refactoring action doesn't add stubs for methods that
are unavailable or not yet introduced.

rdar://32549953

### DIFF
--- a/lib/Sema/SemaDeclObjC.cpp
+++ b/lib/Sema/SemaDeclObjC.cpp
@@ -2133,23 +2133,28 @@ void Sema::CheckImplementationIvars(ObjCImplementationDecl *ImpDecl,
     Diag(IVI->getLocation(), diag::err_inconsistent_ivar_count);
 }
 
-static void WarnUndefinedMethod(Sema &S, SourceLocation ImpLoc,
-                                ObjCMethodDecl *method,
-                                bool &IncompleteImpl,
-                                unsigned DiagID,
-                                NamedDecl *NeededFor = nullptr) {
+static bool shouldWarnUndefinedMethod(const ObjCMethodDecl *M) {
   // No point warning no definition of method which is 'unavailable'.
-  switch (method->getAvailability()) {
+  switch (M->getAvailability()) {
   case AR_Available:
   case AR_Deprecated:
-    break;
+    return true;
 
-      // Don't warn about unavailable or not-yet-introduced methods.
+  // Don't warn about unavailable or not-yet-introduced methods.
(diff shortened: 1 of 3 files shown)


lib/Sema/SemaDeclObjC.cpp
```diff
@@ -2706,7 +2711,9 @@ static void CheckProtocolMethodDefs(
             unsigned DIAG = diag::warn_unimplemented_protocol_method;
             if (!S.Diags.isIgnored(DIAG, ImpLoc)) {
               if (MissingRequirements)
-                HasMissingRequirements = true;
+                HasMissingRequirements =
```
vedantk, line 2714: I think `if (!HasMissingRequirements) ...` would be clearer.

lib/Sema/SemaDeclObjC.cpp
```diff
@@ -2732,7 +2739,8 @@ static void CheckProtocolMethodDefs(
       unsigned DIAG = diag::warn_unimplemented_protocol_method;
       if (!S.Diags.isIgnored(DIAG, ImpLoc)) {
         if (MissingRequirements)
-          HasMissingRequirements = true;
+          HasMissingRequirements =
```
vedantk, line 2742: Ditto.

test/FixIt/fixit-fill-in-protocol-requirements.m
```diff
@@ -52,3 +53,30 @@ @interface FourProtocols <P2, P3, P4> // expected-warning {{class 'ThreeProtocol
 @end
 @implementation FourProtocols
 @end
+
+// Unavailable methods
+@protocol TakeAvailabilityIntoAccount
+
+- (void)unavailableMethod __attribute__((availability(macos,unavailable)));
++ (void)notYetAvailableMethod __attribute__((availability(macos,introduced=10.11)));
+- (void)availableMethod;
```
vedantk, line 62: I think this is missing coverage for deprecated methods?

vedantk (commented):
Mostly looks good!

test/FixIt/fixit-fill-in-protocol-requirements.m
hyp: Right, I'll add it.

hyp:
Thanks, committed to https://github.com/apple/swift-clang
   c790745221..4d7cc27fda  upstream-with-swift -> upstream-with-swift

PR #106, opened 2017-07-17
[refactor] Extract: use substitued types for return types derived from ObjC property expressions

hyp:
Previously extract would produce functions/expressions with unsubstituted ObjC generic argument types.

rdar://33350790


vedantk (approved):
LGTM

PR #110, opened 2017-07-21
[refactor][extract] fix crash

hyp:
We didn't find the begin/end statements in the CS sometimes when selecting ObjC properties because of PseudoObjectExpr

### DIFF
--- a/lib/Tooling/Refactor/Extract.cpp
+++ b/lib/Tooling/Refactor/Extract.cpp
@@ -161,6 +161,22 @@ static bool isMultipleCandidateBinOp(BinaryOperatorKind Op) {
   return Op == BO_Add || Op == BO_Sub;
 }
 
+/// Searches for the selected statement in the given CompoundStatement, looking
+/// through things like PseudoObjectExpressions.
+static CompoundStmt::const_body_iterator
+findSelectedStmt(CompoundStmt::body_const_range Statements,
+                 const Stmt *Target) {
+  return llvm::find_if(Statements, [=](const Stmt *S) {
+    if (S == Target)
+      return true;
+    if (const auto *POE = dyn_cast<PseudoObjectExpr>(S)) {
+      if (POE->getSyntacticForm() == Target)
+        return true;
+    }
+    return false;
+  });
+}
+
 /// Returns the first and the last statements that should be extracted from a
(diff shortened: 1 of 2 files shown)


lib/Tooling/Refactor/Extract.cpp
```diff
@@ -171,8 +184,12 @@ Optional<CompoundStatementRange> getExtractedStatements(const CompoundStmt *CS,
   assert(Begin && End);
   CompoundStatementRange Result;
   Result.First = std::find(CS->body_begin(), CS->body_end(), Begin);
```
vedantk: Is the first std::find necessary? Is there some reason to not "look into" PseudoObjectExpr's on the first pass through the compound stmt?

lib/Tooling/Refactor/Extract.cpp
```diff
@@ -171,8 +184,12 @@ Optional<CompoundStatementRange> getExtractedStatements(const CompoundStmt *CS,
...
+  if (Result.First == CS->body_end())
+    Result.First = comprehensiveFind(CS, Begin);
   assert(Result.First != CS->body_end());
   Result.Last = std::find(Result.First, CS->body_end(), End);
+  if (Result.Last == CS->body_end())
+    Result.Last = comprehensiveFind(CS, End);
```
vedantk: This comment won't apply if you plan on reusing comprehensiveFind elsewhere, but you could avoid some pointer chasing by calling comprehensiveFind once and looking for both Begin & End?

lib/Tooling/Refactor/Extract.cpp
```diff
@@ -161,6 +161,19 @@ static bool isMultipleCandidateBinOp(BinaryOperatorKind Op) {
   return Op == BO_Add || Op == BO_Sub;
 }
 
+static CompoundStmt::const_body_iterator
+comprehensiveFind(const CompoundStmt *CS, const Stmt *Target) {
```
vedantk: Function names should start with a verb. Maybe this could be findTargetsForExtraction?

hyp:
Right, it would be good to merge everything into one search pass! I'll update.

vedantk (approved):
LGTM

PR #111, opened 2017-07-21
Mark the swift_wrapper attribute as inheritable

jrose-apple:
(as it should have been all along)

This prevents a crash in the case where someone happened to redeclare the typedef later, and then different Swift compilation contexts treated the typedef differently.

rdar://problem/31935341

### DIFF
--- a/include/clang/Basic/Attr.td
+++ b/include/clang/Basic/Attr.td
@@ -1711,7 +1711,7 @@ def SwiftName : InheritableAttr {
   let Documentation = [Undocumented];
 }
 
-def SwiftNewtype : Attr {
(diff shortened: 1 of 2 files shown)


jrose-apple:
@adrian-prantl or @devincoughlin, mind reviewing this Clang change for the 4.0 branch, since Michael (and Doug) are both pretty swamped?

adrian-prantl:
Based on my (limited) understanding of clang's AST this change seems safe/reasonable to me.

milseman (approved):
LGTM

jrose-apple:
Thanks, both!

PR #113, opened 2017-07-21
Fix Windows support for Driver and add support to Index

troughton:
This pull request contains fixes to enable Clang to build on Windows against the MSVC toolchain. 

Of particular note is the change from timespec to time_t in IndexDataStore.h; this field doesn't seem to be in use, and time_t is consistent with elsewhere in the repository, but the two types are not equivalent, so I'm not certain this change is correct.

I also if-blocked out some Unix specific APIs around getting the system SDK.

### DIFF
--- a/include/clang/Index/IndexDataStore.h
+++ b/include/clang/Index/IndexDataStore.h
@@ -38,7 +38,7 @@ public:
   struct Event {
     EventKind Kind;
     std::string Filename;
-    timespec ModTime;
+    time_t ModTime;
   };
 
   typedef std::function<void(ArrayRef<Event> Events, bool isInitial)> EventReceiver;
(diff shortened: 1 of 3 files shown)


troughton:
Upon looking more closely at the branching model, I notice I need to rebase off `upstream-with-swift`, in which ToolChains.cpp has been deleted. There may also be other changes necessary for that branch that I haven't yet tested. I'm closing this for now.

omochi:
@troughton 
Hi. I faced build failure on Windows around `timespec` type now.
Will you reopen this PR?

troughton:
Changes moved to new PR https://github.com/apple/swift-clang/pull/127

omochi:
@troughton Thank you 😂

PR #119, opened 2017-09-06
Soft-deprecate -fapinotes-cache-path; use -fmodules-cache-path instead.

jrose-apple:
When Clang decided to change the default module cache path to include the current user's name, the API notes path didn't change. Rather than duplicate that logic, just put the API notes cache in with the module cache by default. We still support -fapinotes-cache-path in case anyone's using it, but from now on -fmodules-cache-path should be sufficient. (This is reasonable because API notes are currently linked to modules anyway.)

[SR-5806](https://bugs.swift.org/browse/SR-5806)


jrose-apple:
Requesting actual review because I plan to cherry-pick this to swift-4.1-branch (even though we're not under the formal review process yet).

DougGregor (approved):
LGTM

PR #120, opened 2017-09-06
On second thought, don't bother caching compiled API notes at all.

jrose-apple:
Because API notes are always used with modules, and modules are always cached in PCM files, we don't really achieve any major wins by caching the API notes information separately. Just rebuild the PCM when the API notes change, like we do for headers.

The next step is to not use the binary form at all. We don't really need it, for the same reason we don't need the cache.

rdar://problem/31745420, also a better solution to [SR-5806](https://bugs.swift.org/browse/SR-5806) than #119.


DougGregor (approved):
My beautiful, beautiful code! You monster! Thank you!

PR #122, opened 2017-09-07
[APINotes] Honor Swift 4 API notes in Swift 3 mode.

jrose-apple:
More generally, change the meaning of the SwiftVersions section to be "this version or earlier" rather than "exactly this version". This keeps us from having to duplicate information for builds of Swift that support more than two possible versions (e.g. 3, 4, and 5).

### DIFF
--- a/lib/APINotes/APINotesReader.cpp
+++ b/lib/APINotes/APINotesReader.cpp
@@ -1487,31 +1487,30 @@ APINotesReader::VersionedInfo<T>::VersionedInfo(
     SmallVector<std::pair<VersionTuple, T>, 1> results)
   : Results(std::move(results)) {
 
-  // Look for an exact version match.
-  Optional<unsigned> unversioned;
-  Selected = Results.size();
+  assert(!Results.empty());
+  assert(std::is_sorted(Results.begin(), Results.end(),
+                        [](const std::pair<VersionTuple, T> &left,
+                           const std::pair<VersionTuple, T> &right) -> bool {
+    assert(left.first != right.first && "two entries for the same version");
+    return left.first < right.first;
+  }));
 
+  Selected = Results.size();
   for (unsigned i = 0, n = Results.size(); i != n; ++i) {
     if (version && Results[i].first >= version) {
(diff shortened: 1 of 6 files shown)


lib/APINotes/APINotesReader.cpp
```diff
@@ -1492,8 +1492,13 @@ APINotesReader::VersionedInfo<T>::VersionedInfo(
...
-    if (Results[i].first == version) {
-      Selected = i;
+    if (version && Results[i].first >= version) {
+      // Pick the closest match. If the current version is "4", then entries for
+      // 4 are better than entries for 5, but both are valid.
+      if (Selected == Results.size() ||
+          Results[Selected].first > Results[i].first) {
+        Selected = i;
+      }
       break;
```
jrose-apple: Oops. This `break` is wrong. Either I need a test to catch this, or I should guarantee that the list is always sorted.

lib/APINotes/APINotesWriter.cpp
```diff
@@ -475,10 +475,16 @@ namespace {
   template<typename T>
   void emitVersionedInfo(
          raw_ostream &out,
-         const SmallVectorImpl<std::pair<VersionTuple, T>> &infoArray,
+         SmallVectorImpl<std::pair<VersionTuple, T>> &infoArray,
          llvm::function_ref<void(raw_ostream &out,
                                  const typename MakeDependent<T>::Type& info)>
            emitInfo) {
+    std::sort(infoArray.begin(), infoArray.end(),
```
DougGregor, line 482: No `array_pod_sort`, eh?

DougGregor (approved):
LGTM, nice testing.

lib/APINotes/APINotesWriter.cpp
jrose-apple: It's not necessarily POD, is it?

DougGregor: Well. Maybe not by the language definition., but it'd work. `std::sort()` is not exactly a small thing to instantiate.

PR #123, opened 2017-09-09
[APINotes] Record what version caused an annotation to get replaced.

jrose-apple:
(Depends on #122; this PR is intended for just the third commit.)

Now that -fapinotes-swift-version=3 can pick up API notes for "Version: 4", the information for *inactive* API notes isn't sufficient to tell what would have happened under a different version. That is, I can't ask "what would happen under version 4", because I can't tell if the active annotations are active because they were specified for version 3 or version 4. The inactive, unversioned attributes were no help, either: they just said "version 0".

Fix this by adding a new flag 'IsReplacedByActive' to SwiftVersionedAttr and SwiftVersionedRemovalAttr. When set, the 'Version' field refers to the API note that caused this annotation to become inactive.

Note that "version 0" can still exist: when an attribute written in source is replaced by an unversioned attribute in the API notes.

### DIFF
--- a/include/clang/Basic/Attr.td
+++ b/include/clang/Basic/Attr.td
@@ -1746,7 +1746,8 @@ def SwiftVersioned : Attr {
   // This attribute has no spellings as it is only ever created implicitly
   // from API notes.
   let Spellings = [];
-  let Args = [VersionArgument<"Version">, AttrArgument<"AttrToAdd">];
+  let Args = [VersionArgument<"Version">, AttrArgument<"AttrToAdd">,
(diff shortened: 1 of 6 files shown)


jrose-apple:
I think I'm going to have to make this an option, sadly, or else we're going to have a Clang/Swift revlock.

(where the Swift change is non-trivial)

jrose-apple:
…ah, well, maybe that's acceptable. update-checkout all the working copies!

DougGregor (approved):
Yeah, seems reasonable.

PR #127, opened 2017-09-19
Fix Windows support for Driver and add support to Index

troughton:
This pull request contains fixes to enable Clang to build on Windows against the MSVC toolchain.

Of particular note is the change from timespec to time_t in IndexDataStore.h; this field doesn't seem to be in use, and time_t is consistent with elsewhere in the repository, but the two types are not equivalent, so I'm not certain this change is correct.

https://github.com/apple/swift-clang/pull/113 rebased off upstream-with-swift

### DIFF
--- a/include/clang/Index/IndexDataStore.h
+++ b/include/clang/Index/IndexDataStore.h
@@ -38,7 +38,7 @@ public:
   struct Event {
     EventKind Kind;
     std::string Filename;
-    timespec ModTime;
+    time_t ModTime;
   };
 
   typedef std::function<void(ArrayRef<Event> Events, bool isInitial)> EventReceiver;
@@ -71,7 +71,7 @@ public:
   struct UnitEvent {
     UnitEventKind Kind;
     StringRef UnitName;
-    timespec ModTime;
+    time_t ModTime;
   };
   struct UnitEventNotification {
     bool IsInitial;
--- a/include/clang/Index/IndexDataStore.h
+++ b/include/clang/Index/IndexDataStore.h
@@ -14,6 +14,7 @@
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/Chrono.h"
 #include <functional>
 #include <memory>
 #include <string>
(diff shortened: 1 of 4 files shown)


akyrtzi:
It *is* getting used, see function indexstore_unit_event_get_modification_time, part of the libIndexStore C library. That function should not change, it would break the ABI.
Also it originates from `DirectoryWatcher.cpp`, I'm not sure why you didn't have to touch that or the C library.

`timespec` is supposed to be part of `time.h`, does the Windows headers really not provide it ?

troughton:
`timespec` isn't included on Windows; a quick web search reveals a number of different suggestions for replacement that all differ slightly in functionality.

I missed DirectoryWatcher.cpp because I (mistakenly) only tested on Windows, where that code is hidden behind a `HAVE_CORESERVICES` define, so the type wasn't actually getting used.

Maybe a better solution would be just to declare the `timespec` struct on Windows; there's nothing in it that's inherently incompatible. If I were to do that, where would be the best place to declare it – in that same header file?

akyrtzi:
We could replace it in the C++ code with llvm's TimePoint (from `llvm/Support/Chrono.h`), `getLastModificationTime()` uses that and DirectoryWatcher converts from it, but it could just pass it directly.
Then keep timespec in the C library and convert from TimePoint with the same way as DirectoryWatcher does it.

troughton:
@akyrtzi Were you thinking something along the lines of these changes?

akyrtzi:
You can also remove the `toTimeSpec()` function from `DirectoryWatcher.cpp`.
Also you can use `SafelyCloseFileDescriptor()` from `llvm/Support/Process.h` and avoid the platform-specific code in `IndexUnitReader.cpp`.

Otherwise LGTM.

troughton:
Done, thanks.

akyrtzi:
Is this part
```
+#if defined(_WIN32)
+#include <io.h>
+#else 
```
in `IndexUnitReader.cpp` still necessary ?

troughton:
No, it's not; I got rid of it, and I also removed the `unistd.h` include since that's also now unnecessary.

akyrtzi (approved):

troughton:
Just checking in - are we okay to merge this now?

akyrtzi:
@swift-ci smoke test and merge

akyrtzi:
@swift-ci Please test

akyrtzi:
@swift-ci Please test

akyrtzi:
master-next branch has issues, but I tested the changes on swift-4.1 and they look good, merging.

PR #129, opened 2017-09-27
[Atomic][X8664] set max atomic inline width according to the target

dcci:
This merges a fix to build swift on FreeBSD. Hopefully I did all the steps correctly, and if not, the bot will yell at me :)

This is to fix PR31620. MaxAtomicInlineWidth is set to 128 for x86_64. However
for target without cx16 support, 128 atomic operation will generate __sync_*
libcalls. The patch set MaxAtomicInlineWidth to 64 if the target doesn't support
cx16.

Differential Revision: https://reviews.llvm.org/D38046

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@313992 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/include/clang/Basic/TargetInfo.h
+++ b/include/clang/Basic/TargetInfo.h
@@ -447,6 +447,9 @@ public:
   /// \brief Return the maximum width lock-free atomic operation which can be
   /// inlined given the supported features of the given target.
   unsigned getMaxAtomicInlineWidth() const { return MaxAtomicInlineWidth; }
+  /// \brief Set the maximum inline or promote width lock-free atomic operation
+  /// for the given target.
+  virtual void setMaxAtomicWidth() {}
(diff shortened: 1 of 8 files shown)


dcci:
@bob-wilson @fredriss can you please take a look?

bob-wilson:
Two big comments:

1. The "stable" branch is locked -- it is just an alias for the latest release branch, which is currently swift-4.1-branch. That's where this change should go.

2. The lib/Basic/Targets/ reorganization in r308791 isn't in the 4.1 branch, so you need to fix the patch to apply in the right place with the old organization. The current change here is pulling in 1,000+ lines in the X86.h file that doesn't otherwise exist on this branch.

dcci:
@bob-wilson sure, thanks for your help :). I'll close this one and open a new pull request.

PR #130, opened 2017-09-27
[index] Fix shared libs build breakage: clangIndex and clangFrontend were mutually dependent

nathawes:
Resolves rdar://problem/33603328.

### DIFF
--- a/include/clang/Frontend/CompilerInstance.h
+++ b/include/clang/Frontend/CompilerInstance.h
@@ -183,6 +183,12 @@ class CompilerInstance : public ModuleLoader {
   /// The list of active output files.
   std::list<OutputFile> OutputFiles;
 
+  /// \brief An optional callback function used to wrap all FrontendActions
+  /// produced to generate imported modules before they are executed.
+  std::function<std::unique_ptr<FrontendAction>
+    (const FrontendOptions &opts, std::unique_ptr<FrontendAction> action)>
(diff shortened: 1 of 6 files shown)


include/clang/Frontend/CompilerInstance.h
```diff
@@ -183,6 +183,12 @@ class CompilerInstance : public ModuleLoader {
   /// The list of active output files.
   std::list<OutputFile> OutputFiles;
 
+  /// \brief An optional callback function used to wrap all FrontendActions
+  /// produced to generate imported modules before they are executed.
+  std::function<std::unique_ptr<FrontendAction>
+    (FrontendOptions &opts, std::unique_ptr<FrontendAction> action)>
```
akyrtzi: Change 'FrontendOptions &opts' to 'const FrontendOptions &opts', to disallow the callee from changing the options.

lib/FrontendTool/ExecuteCompilerInvocation.cpp
```diff
@@ -168,6 +168,7 @@ CreateFrontendAction(CompilerInstance &CI) {
   if (!FEOpts.IndexStorePath.empty()) {
 #if defined(__APPLE__)
```
akyrtzi: This preprocessor check is no longer necessary, right ?

akyrtzi (changes requested):

PR #137, opened 2017-10-26
[CGBlocks] Improve line info in backtraces containing *_helper_block

vedantk:
Instead of only setting a non-zero debug location on the return
instruction in *_helper_block functions, set a proper location on all
instructions within these functions. Pick the start location of the
block literal expr for maximum clarity.

The debugger does not step into *_helper_block functions during normal
single-stepping because we mark their parameters as artificial. This is
what we want (the functions are implicitly generated and uninteresting
to most users). The stepping behavior is unchanged by this patch.

rdar://32907581

Differential Revision: https://reviews.llvm.org/D39310

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@316704 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 290485ca2ff922bb3da08c784323e7e25acbcfde)

### DIFF
--- a/lib/CodeGen/CGBlocks.cpp
+++ b/lib/CodeGen/CGBlocks.cpp
@@ -1530,10 +1530,8 @@ CodeGenFunction::GenerateCopyHelperFunction(const CGBlockInfo &blockInfo) {
 
   CGM.SetInternalFunctionAttributes(nullptr, Fn, FI);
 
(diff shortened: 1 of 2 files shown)


vedantk:
@swift-ci Please test OS X platform

PR #139, opened 2017-11-13
[rename] Find ObjC class textual occurrences in string literals

hyp:
Occurrences match only when the string's text is exactly the same as class name.

rdar://33861206

### DIFF
--- a/lib/Tooling/Refactor/RenameIndexedFile.cpp
+++ b/lib/Tooling/Refactor/RenameIndexedFile.cpp
@@ -14,6 +14,7 @@
 #include "clang/Basic/SourceManager.h"
 #include "clang/Frontend/CompilerInstance.h"
 #include "clang/Frontend/FrontendActions.h"
+#include "clang/Lex/LiteralSupport.h"
 #include "clang/Lex/Preprocessor.h"
 #include "clang/Tooling/Refactor/RefactoringOptions.h"
 #include "llvm/ADT/STLExtras.h"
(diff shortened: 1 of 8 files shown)


lib/Tooling/Refactor/RenameIndexedFile.cpp
```diff
@@ -333,12 +372,18 @@ static void findMatchingTextualOccurrences(
...
         CommentMatches.clear();
       }
-    } else if (!SelectorParsers.empty()) {
+    } else if (ScanNonCommentTokens) {
       for (auto &Parser : SelectorParsers) {
         if (Parser.handleToken(RawTok))
           MatchHandler(OldSymbolOccurrence::MatchingSelector,
                        Parser.SelectorLocations, Parser.SymbolIndex);
       }
+      for (auto &Parser : StringParsers) {
```
benlangmuir, line 381: Why is StringParser `checkToken` while the selector parser uses `handleToken`?

benlangmuir (commented):
What's your reasoning for only doing this for ObjC class names?  I'm okay with starting with the narrow case, but I'm curious what the motivation was there.

hyp:
I think that class names are used in this kind of pattern more often than other declarations. Perhaps this happens with selectors too, but in general `@selector` is more common. Expanding this to other declarations would probably create too much noise, don't you think?

lib/Tooling/Refactor/RenameIndexedFile.cpp
hyp: Fixed

benlangmuir:
After some thought, I think this is the right starting place.  We could fairly reliably search for selector names in strings except for the case of zero-argument selectors, but it would be surprising to have them behave differently than other selectors.  If we find a strong need in the future, we can think about extending this then.

benlangmuir (approved):

PR #140, opened 2017-11-17
[Modules] Change private modules rules and warnings

bcardosolopes:
We used to advertise private modules to be declared as submodules
(Foo.Private). This has proven to not scale well since private headers
might carry several dependencies, introducing unwanted content into the
main module and often causing dep cycles.

Change the canonical way to name it to Foo_Private, forcing private
modules as top level ones, and provide warnings under -Wprivate-module
to suggest fixes for other private naming. Update documentation to
reflect that.

-Wprivate-module is now ignored by default (allowing time for transition
from Foo.Private without breaking all users) and is checked for before
emitting the diagnostics.

rdar://problem/31173501


bcardosolopes:
@swift-ci please test

jrose-apple (commented):
I skimmed it and didn't see anything obviously wrong, but please include some tests for both `@import A.Private` and `@import A_Private` with and without having imported `A`.

bcardosolopes:
@jrose-apple updated with the requested additional tests

jrose-apple:
Ah, specifically we wanted `@import A.Private` to be able to find `A_Private`, right? I don't see a test for that.

bcardosolopes:
@jrose-apple right, but this isn't implemented in this PR, this is "remove old warning behavior and enforce new policy" part of the work.

Remaping `A.private` to `A_Private` in case the former doesn't match anything should come next.

PR #141, opened 2017-11-27
r319002

compnerd:


compnerd:
CC: @bob-wilson 

This will be needed with the next merge

jrose-apple:
cc @vsapsai

vsapsai:
I've fixed r319002 consequences in b94d53a8b4f3c6be05e039a2816939882b5fda85 and ab58a5c1c7b53f3b4008b07262fc19faa99f377a. Build bots are happy and I don't see any other places that have to be updated. But I can be wrong and another pair of eyes can help with spotting more areas to fix.

PR #142, opened 2017-12-06
Unblock SwiftCC usage on Windows

troughton:
We don't have a sensible way to mangle functions using the Swift calling convention on Windows currently. For now, as per [discussions on swift-dev](https://lists.swift.org/pipermail/swift-dev/Week-of-Mon-20171204/006123.html), don't do anything special and mangle it like the C calling convention.

Also in this patch, address [feedback](https://reviews.llvm.org/D31372) on usage of `llvm_unreachable`.

### DIFF
--- a/lib/AST/MicrosoftMangle.cpp
+++ b/lib/AST/MicrosoftMangle.cpp
@@ -2131,9 +2131,12 @@ void MicrosoftCXXNameMangler::mangleCallingConvention(CallingConv CC) {
 
   switch (CC) {
     default:
-      llvm_unreachable("Unsupported CC for mangling");
+      llvm::errs() << "Unsupported CC for mangling: " << CC << ".\n";
     case CC_Win64:
     case CC_X86_64SysV:
+    // FIXME: SwiftCC should have its own mangling specifier.
+    // For now, don't do anything special for SwiftCC mangling.
+    case CC_Swift:
     case CC_C: Out << 'A'; break;
     case CC_X86Pascal: Out << 'C'; break;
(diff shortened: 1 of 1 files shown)


lib/AST/MicrosoftMangle.cpp
```diff
@@ -2131,9 +2131,12 @@ void MicrosoftCXXNameMangler::mangleCallingConvention(CallingConv CC) {
 
   switch (CC) {
     default:
-      llvm_unreachable("Unsupported CC for mangling");
+      llvm::errs() << "Unsupported CC for mangling: " << CC << ".\n";
     case CC_Win64:
     case CC_X86_64SysV:
+    // FIXME: SwiftCC should have its own mangling specifier.
+    // For now, don't do anything special for SwiftCC mangling.
+    case CC_Swift:
```
compnerd, line 2139: This will add `A` to the mangling.

compnerd (changes requested):

lib/AST/MicrosoftMangle.cpp
troughton: Is there a better option from the preexisting symbols? The idea here is just to treat SwiftCC like any other standard calling convention. What would you suggest?

rjmccall: I think A is fine for now.  But (1) this should go in LLVM, not in our private fork, and (2) when you bring it up there, someone might point out a friendlier way of doing it.

troughton:
Submitted as a patch to Clang upstream at https://reviews.llvm.org/D40929.

PR #148, opened 2017-12-16
[CodeGen] Specialize mixed-sign mul-with-overflow (fix PR34920)

vedantk:
This patch introduces a specialized way to lower overflow-checked
multiplications with mixed-sign operands. This fixes link failures and
ICEs on code like this:

  void mul(int64_t a, uint64_t b) {
    int64_t res;
    __builtin_mul_overflow(a, b, &res);
  }

The generic checked-binop irgen would use a 65-bit multiplication
intrinsic here, which requires runtime support for _muloti4 (128-bit
multiplication), and therefore fails to link on i386. To get an ICE
on x86_64, change the example to use __int128_t / __uint128_t.

Adding runtime and backend support for 65-bit or 129-bit checked
multiplication on all of our supported targets is infeasible.

This patch solves the problem by using simpler, specialized irgen for
the mixed-sign case.

llvm.org/PR34920, rdar://34963321

Testing: Apart from check-clang, I compared the output from this fairly
comprehensive test driver using unpatched & patched clangs:
https://gist.github.com/vedantk/3eb9c88f82e5c32f2e590555b4af5081

Differential Revision: https://reviews.llvm.org/D41149

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@320902 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 87961dfbc9f2faf85ccc5342a8415500938ed24a)

### DIFF
--- a/lib/CodeGen/CGBuiltin.cpp
+++ b/lib/CodeGen/CGBuiltin.cpp
@@ -837,6 +837,93 @@ RValue CodeGenFunction::emitBuiltinOSLogFormat(const CallExpr &E) {
   return RValue::get(BufAddr.getPointer());
 }
 
+/// Determine if a binop is a checked mixed-sign multiply we can specialize.
+static bool isSpecialMixedSignMultiply(unsigned BuiltinID,
+                                       WidthAndSignedness Op1Info,
+                                       WidthAndSignedness Op2Info,
+                                       WidthAndSignedness ResultInfo) {
+  return BuiltinID == Builtin::BI__builtin_mul_overflow &&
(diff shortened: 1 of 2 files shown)


vedantk:
@swift-ci Please smoke test and merge

vedantk:
@swift-ci Please smoke test OS X platform

vedantk:
Hi @shahmishal, have I gotten the CI incantations right here? Is there a way to smoke test & merge PR's to swift-clang?

vedantk:
The bots are ignoring me, but local testing shows that this is clean so I'll go ahead and merge.

PR #149, opened 2017-12-20
Revert "[driver][darwin] Take the OS version specified in "-target" a…

xedin:
…s the target"

This reverts commit 1ac71826854a809f67ae15ddacaaa9a6eb8189e8.


xedin:
@hyp Looks like you forgot to revert it from stable as well, which leaves Swift master testing broken :(

hyp:
I think you need to revert 7e5e5284690d0fd246f88a364c73caba16ac55ed instead (there is another follow up commit, otherwise there'll be merge conflicts)

hyp:
I manually merged in https://github.com/apple/swift-clang/pull/150

PR #151, opened 2017-12-21
[Modules] Map private framework submodules to new syntax

bcardosolopes:
Map private framework submodules from Foo.Private to Foo_Private

In case submodule Foo.Private do not exist, try to match Foo_Private,
which is the current suggested way to write private framework modules.

Emit warnings to that users can use the recommended `@import Foo_Private`
form.

Example:

$ cat test.m                                                                                                                                                                                                                                                           
@import Foo.Private;
int a = MYPRIVATE;

$ clang ...
test.m:1:13: warning: no submodule named 'Private' in module 'Foo'; using top level 'Foo_Private' [-Wprivate-module]
@import Foo.Private;
        ~~~~^~~~~~~
        Foo_Private
./Foo.framework/Modules/module.private.modulemap:2:18: note: module defined here
framework module Foo_Private {
                 ^
1 warning generated.

rdar://problem/36023940

### DIFF
--- a/include/clang/Basic/DiagnosticFrontendKinds.td
+++ b/include/clang/Basic/DiagnosticFrontendKinds.td
@@ -196,6 +196,11 @@ def err_missing_module : Error<
 def err_no_submodule : Error<"no submodule named %0 in module '%1'">;
 def err_no_submodule_suggest : Error<
   "no submodule named %0 in module '%1'; did you mean '%2'?">;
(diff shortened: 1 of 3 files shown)


bcardosolopes:
@swift-ci please test

jrose-apple:
We'll have to diagnose this specially on the Swift side too, since we don't have real source locations for module includes, but it's okay, we know how to do that. This seems reasonable!

PR #152, opened 2018-01-17
Make indexstore.h importable by Swift.

allevato:
This change removes some C++-specific things that prevented Swift from being able to import this header (when provided with a module map).


jrose-apple:
Ah, please get this committed at llvm.org first! Then we can cherry-pick it into either a release branch or just into `stable` for holding. For more info see swift/docs/Branches.md.

allevato:
@jrose-apple Thanks for pinging this! I originally put this PR here because the indexstore API hadn't yet been accepted into mainline clang (https://reviews.llvm.org/D39050); unfortunately, even though it's been merged now, it's in a completely different form that's not compatible with this change so I'll abandon it and revisit what I was working on.

PR #153, opened 2018-01-26
[index] Fix crash when indexing a C++14 PCH/module related to TemplateTemplateParmDecls of alias templates

akyrtzi:
TemplateTemplateParmDecls of alias templates ended-up serialized as 'file-level decls' which was causing a crash while trying to index a PCH/module file that contained them.
Commit makes sure TemplateTemplateParmDecls are not recorded as such kind of decls.

Fixes crash of rdar://36608297

### DIFF
--- a/lib/Index/IndexDecl.cpp
+++ b/lib/Index/IndexDecl.cpp
@@ -665,8 +665,11 @@ public:
 
   bool VisitTemplateDecl(const TemplateDecl *D) {
 
-    // Index the default values for the template parameters.
     const NamedDecl *Parent = D->getTemplatedDecl();
+    if (!Parent)
+      return true;
+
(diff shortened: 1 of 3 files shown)


akyrtzi:
* Explanation: There is a clang index-while-building crash when indexing the SDK with C++ modules enabled. TemplateTemplateParmDecls of alias templates ended-up serialized as 'file-level decls' which was causing a crash while trying to index a PCH/module file that contained them.
Commit makes sure TemplateTemplateParmDecls are not recorded as such kind of decls.
* Risk: Low, simple fix, only affects index-while-building path.
* Reviewed By: Nathan Hawes
* Testing: Automated test suite with extra cases
* Directions for QA: N/A
* Radar: rdar://36608297

PR #160, opened 2018-02-05
ppc64le port.

asowani:
This file contains changes required to build and use it on ppc64le platform.

### DIFF
--- a/lib/Basic/Targets.cpp
+++ b/lib/Basic/Targets.cpp
@@ -1777,6 +1777,15 @@ public:
     }
     return false;
   }
+
+  CallingConvCheckResult checkCallingConvention(CallingConv CC) const override {
+    switch (CC) {
+      case CC_Swift:
+        return CCCR_OK;
+      default:
+        return CCCR_Warning;
+    }
+  }
 };
 
 class DarwinPPC32TargetInfo : public DarwinTargetInfo<PPC32TargetInfo> {
--- a/lib/Basic/Targets.cpp
+++ b/lib/Basic/Targets.cpp
@@ -1780,10 +1780,10 @@ public:
 
   CallingConvCheckResult checkCallingConvention(CallingConv CC) const override {
     switch (CC) {
-      case CC_Swift:
-        return CCCR_OK;
-      default:
-        return CCCR_Warning;
+    case CC_Swift:
+      return CCCR_OK;
+    default:
+      return CCCR_Warning;
     }
   }
 };
--- a/lib/CodeGen/TargetInfo.cpp
+++ b/lib/CodeGen/TargetInfo.cpp
@@ -4189,7 +4189,7 @@ PPC32TargetCodeGenInfo::initDwarfEHRegSizeTable(CodeGen::CodeGenFunction &CGF,
 
 namespace {
 /// PPC64_SVR4_ABIInfo - The 64-bit PowerPC ELF (SVR4) ABI information.
-class PPC64_SVR4_ABIInfo : public ABIInfo {
+class PPC64_SVR4_ABIInfo : public SwiftABIInfo {
 public:
   enum ABIKind {
     ELFv1 = 0,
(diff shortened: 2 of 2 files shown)


asowani:
These code changes need to go in along with following PRs:
https://github.com/apple/swift-corelibs-foundation/pull/1421 (merged)
https://github.com/apple/swift-package-manager/pull/1482 (cleared for merge)
https://github.com/apple/swift/pull/14407 (to be merged)

lib/Basic/Targets.cpp
```diff
@@ -1777,6 +1777,15 @@ class PPC64TargetInfo : public PPCTargetInfo {
     }
     return false;
   }
+
+  CallingConvCheckResult checkCallingConvention(CallingConv CC) const override {
+    switch (CC) {
+      case CC_Swift:
```
xwu: LLVM style is to have `case` aligned with `switch`—no indenting.

lib/CodeGen/TargetInfo.cpp
```diff
@@ -4276,6 +4276,15 @@ class PPC64_SVR4_ABIInfo : public ABIInfo {
 
   Address EmitVAArg(CodeGenFunction &CGF, Address VAListAddr,
                     QualType Ty) const override;
+
+  bool shouldPassIndirectlyForSwift( ArrayRef<llvm::Type*> scalars,
```
xwu: Prevailing style in this project is UpperCamelCase variables. Also, there’s an extra space after the opening paren.

xwu (commented):
Just some drive-by style nits.

asowani:
@xwu I have corrected styling issues. Thanks for pointing out!

jrose-apple:
Hi, Atul! Committing things to Clang and LLVM gets tricky around Swift because of how the branches are used. In particular, I'm going to ask you to commit this to the `upstream-with-swift` branch first, and then cherry-pick it to `stable` and ~`clang-5.0-branch`~ `swift-5.0-branch`. I suggest reading [docs/Branches.md](https://github.com/apple/swift/blob/master/docs/Branches.md) in the Swift repository to make sense of this.

bob-wilson:
Good catch! Jordan is absolutely right that this should go to upstream-with-swift first. For cherry-picking, it should definitely go to swift-5.0-branch (slightly different branch name than what Jordan said). The stable branch is more tricky. We generally keep that branch locked and use it as a "branch alias" for the current release branch, which is swift-4.1-branch. There is an auto merger that keeps stable updated. Since we're in the last stages of work on Swift 4.1, I would generally prefer not to put these changes there. We're planning to switch over to develop Swift against the newer swift-5.0-branch in the next few weeks. How inconvenient would it be to wait until then?

bob-wilson:
If it would be a hassle, please say so. I'm really just being paranoid here -- these changes shouldn't add any risk to other platforms.

asowani:
@bob-wilson I can definitely wait till 5.0 branch switch-over happens. Does this apply to all related PRs I have raised in other swift repositories as well?

bob-wilson:
If those other PRs can go in separately, that's fine. If they need to happen at the same time as the Clang/LLVM change, then you should wait.

In any case, you can go ahead with PRs for upstream-with-swift and swift-5.0-branch now.

asowani:
@jrose-apple @bob-wilson I checked out upstream-with-swift branch (using git checkout upstream-with-swift) but the version of files there seem too different. For example, lib/Basic/Targets.cpp file hardly match. Am I looking at correct branch?

bob-wilson:
I think so. Compare that branch vs. the upstream llvm.org Clang sources. As I recall, the target code was refactored into multiple files since we created the swift-4.1-branch last summer.

asowani:
As discussed, I have merged these changes with "upstream-with-swift" branch here: https://github.com/apple/swift-clang/pull/167

asowani:
@jrose-apple @bob-wilson Any idea when this PR will be merged? Thanks!

bob-wilson:
This PR for stable should not be merged. After the change is in the upstream-with-swift branch, it can be cherry-picked to swift-4.2-branch (which is currently tracked by stable).

PR #161, opened 2018-02-09
Tooling: use /bigobj when building a file

compnerd:
This file has enough symbols that it extends beyond the PE/COFF
specification.  Use `/bigobj` to enable builds on Windows.


compnerd:
CC: @bob-wilson

bob-wilson:
This looks fine to me.

PR #164, opened 2018-02-22
[WIP] [APINotes] Add a 'swift_nonbridged' attribute for typedefs

jrose-apple:
Some APIs really need reference semantics even for a type that's normally bridged to a Swift value type, such as the NSString view onto NSTextStorage's backing in AppKit. Up until now people have had to work around this with hacks, usually wrapper functions that return an "NSObject" that then can be downcast back to NSString. This new Clang attribute avoids that.

rdar://problem/28480421

* * *

This PR is targeting the `stable` branch, and should not be merged until this change is also in `upstream-with-swift`.


jrose-apple:
Tony provided some reasons why a typedef was a limiting approach. I'll go with a type attribute instead.

PR #165, opened 2018-02-27
APINotes: adjust for upstream SVN r326091

compnerd:
`HashString` was replaced in favour of `djbHash`.  Note that the seed
for the hashing has changed.


compnerd:
CC: @fredriss @bob-wilson

bob-wilson:
This is identical to the change that I was testing locally today.

PR #173, opened 2018-04-03
CodeGenCXX: support PreserveMostCC in MS ABI

compnerd:
Microsoft has reserved 'U' for the PreserveMostCC which is used in the
swift runtime.  Add support for this.  This allows the swift runtime to
be built for Windows again.

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@329025 91177308-0d34-0410-b5e6-96231b3b80d8


compnerd:
CC: @tkremenek @bob-wilson 

This is needed to build for Windows, as this calling convention was not previously supported in the MS ABI mode.  It is just a back port of an upstream'ed change.

tkremenek:
@swift-ci test

tkremenek:
@compnerd this is already in `upstream-with-swift`?

compnerd:
@tkremenek, yeah, its in upstream-with-swift as it came from llvm.org, the nightly merges picked it up.   Thanks!

PR #174, opened 2018-04-09
upstream-update: adjust for SVN r329570

compnerd:
The interface was changed to return SourceLocation to consumers, not
FileID/Offset pair.  Adjust the overload accordingly.


compnerd:
lol, @akyrtzi beat me by a few minutes.

PR #176, opened 2018-04-12
[Modules] Use export_as for autolinking frameworks

bcardosolopes:
framework module SomeKitCore {
  ...
  export_as SomeKit
}

Given the module above, while generting autolink information during
codegen, clang should to emit '-framework SomeKitCore' only if SomeKit
was not imported in the relevant TU, otherwise it should use '-framework
SomeKit' instead.

rdar://problem/38269782

### DIFF
--- a/include/clang/Basic/Module.h
+++ b/include/clang/Basic/Module.h
@@ -334,6 +334,10 @@ public:
   /// an entity from this module is used.
   llvm::SmallVector<LinkLibrary, 2> LinkLibraries;
 
+  /// \brief Autolinking uses the framework name for linking purposes
+  /// when this is false and the export_as name otherwise.
+  bool UseExportAsModuleLinkName = false;
+
   /// \brief The set of "configuration macros", which are macros that
   /// (intentionally) change how this module is built.
   std::vector<std::string> ConfigMacros;
--- a/include/clang/Lex/ModuleMap.h
+++ b/include/clang/Lex/ModuleMap.h
@@ -21,6 +21,7 @@
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/PointerIntPair.h"
+#include "llvm/ADT/StringSet.h"
 #include "llvm/ADT/SmallPtrSet.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringMap.h"
@@ -112,7 +113,19 @@ class ModuleMap {
   /// \brief The number of modules we have created in total.
   unsigned NumCreatedModules = 0;
 
+  /// In case a module has a export_as entry, it might have a pending link
+  /// name to be determined if that module is imported.
+  llvm::StringMap<llvm::StringSet<>> PendingLinkAsModule;
+
 public:
+  /// Use PendingLinkAsModule information to mark top level link names that
+  /// are going to be replaced by export_as aliases.
+  void resolveLinkAsDependencies(Module *Mod);
+
+  /// Make module to use export_as as the link dependency name if enough
+  /// information is available or add it to a pending list otherwise.
+  void addLinkAsDependency(Module *Mod);
+
   /// \brief Flags describing the role of a module header.
   enum ModuleHeaderRole {
     /// \brief This header is normally included in the module.
--- a/lib/CodeGen/CodeGenModule.cpp
+++ b/lib/CodeGen/CodeGenModule.cpp
@@ -1328,6 +1328,12 @@ static void addLinkOptionsPostorder(CodeGenModule &CGM, Module *Mod,
   // Add linker options to link against the libraries/frameworks
   // described by this module.
   llvm::LLVMContext &Context = CGM.getLLVMContext();
+
+  // For modules that use export_as for linking, use that module
+  // name instead.
+  if (Mod->UseExportAsModuleLinkName)
+    return;
+
   for (unsigned I = Mod->LinkLibraries.size(); I > 0; --I) {
     // Link against a framework.  Frameworks are currently Darwin only, so we
     // don't to ask TargetCodeGenInfo for the spelling of the linker option.
--- a/lib/Frontend/CompilerInstance.cpp
+++ b/lib/Frontend/CompilerInstance.cpp
@@ -2014,6 +2014,12 @@ CompilerInstance::loadModule(SourceLocation ImportLoc,
                      Module, ImportLoc);
   }
 
+  // Resolve any remaining module using export_as for this one.
+  getPreprocessor()
+      .getHeaderSearchInfo()
+      .getModuleMap()
+      .resolveLinkAsDependencies(TopModule);
+
   LastModuleImportLoc = ImportLoc;
   LastModuleImportResult = ModuleLoadResult(Module);
   return LastModuleImportResult;
--- a/lib/Lex/ModuleMap.cpp
+++ b/lib/Lex/ModuleMap.cpp
@@ -54,6 +54,24 @@
 
 using namespace clang;
 
+void ModuleMap::resolveLinkAsDependencies(Module *Mod) {
+  auto PendingLinkAs = PendingLinkAsModule.find(Mod->Name);
+  if (PendingLinkAs != PendingLinkAsModule.end()) {
+    for (auto &Name : PendingLinkAs->second) {
+      auto *M = findModule(Name.getKey());
+      if (M)
+        M->UseExportAsModuleLinkName = true;
+    }
+  }
+}
+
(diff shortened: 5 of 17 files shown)


include/clang/Basic/Module.h
```diff
@@ -334,6 +334,10 @@ class Module {
   /// an entity from this module is used.
   llvm::SmallVector<LinkLibrary, 2> LinkLibraries;
 
+  /// \brief The set of libraries or frameworks to link against when
+  /// an entity from this module is used.
```
jrose-apple: This comment never got updated.

lib/CodeGen/CodeGenModule.cpp
```diff
@@ -1332,9 +1332,16 @@ static void addLinkOptionsPostorder(CodeGenModule &CGM, Module *Mod,
     // Link against a framework.  Frameworks are currently Darwin only, so we
     // don't to ask TargetCodeGenInfo for the spelling of the linker option.
     if (Mod->LinkLibraries[I-1].IsFramework) {
+      StringRef LibName = Mod->LinkLibraries[I - 1].Library;
+
+      // For modules that use export_as for linking, use that module
+      // name instead.
```
jrose-apple: It seems like we should support `export_as` with non-frameworks as well, but maybe that's not important right now.

include/clang/Lex/ModuleMap.h
```diff
@@ -112,7 +113,15 @@ class ModuleMap {
   /// \brief The number of modules we have created in total.
   unsigned NumCreatedModules = 0;
 
+  /// In case a module has a export_as entry, it might have a pending link
+  /// name to be determined if that module is imported.
+  llvm::StringMap<llvm::StringSet<>> PendingLinkAsModule;
```
jrose-apple, line 118: I suggest using `TinyPtrVector<Module *>` rather than `StringSet`.

lib/CodeGen/CodeGenModule.cpp
```diff
@@ -1332,9 +1332,16 @@ static void addLinkOptionsPostorder(CodeGenModule &CGM, Module *Mod,
...
+      if (Mod->Name == LibName && Mod->UseExportAsModuleLinkName)
```
jrose-apple: Why limit this to `Mod->Name == LibName`? Is this to handle private modules?

lib/Frontend/CompilerInstance.cpp
```diff
@@ -2014,6 +2014,12 @@ CompilerInstance::loadModule(SourceLocation ImportLoc,
                      Module, ImportLoc);
   }
 
+  // Resolve any remaining module using export_as for this one.
+  getPreprocessor()
+      .getHeaderSearchInfo()
+      .getModuleMap()
+      .resolveLinkAsDependencies(TopModule);
```
jrose-apple, line 2021: This is already called in ASTReader. Does it need to be called again?

lib/Lex/ModuleMap.cpp
```diff
@@ -54,6 +54,17 @@
...
 
+void ModuleMap::resolveLinkAsDependencies(Module *Mod) {
+  auto PendingLinkAs = PendingLinkAsModule.find(Mod->Name);
+  if (PendingLinkAs != PendingLinkAsModule.end()) {
+    for (auto &Name : PendingLinkAs->second) {
+      auto *M = findModule(Name.getKey());
+      if (M)
+        M->UseExportAsModuleLinkName = true;
+    }
+  }
```
jrose-apple, line 65: Maybe remove the entry after this loop is done? That could keep the table from having to grow (possibly).

test/Modules/use-exportas-for-link.m
```diff
@@ -0,0 +1,44 @@
...
+// CHECK_E: ![[MODULE]] = !{!"-framework", !"SomeKitCore"}
+#ifdef E
+@import OtherKit;
+#endif
+
+// RUN: %clang_cc1 -emit-llvm -o - -fmodules-cache-path=%t -DF -fmodules -fimplicit-module-maps -F %S/Inputs/exportas-link %s | FileCheck --check-prefix=CHECK_F %s
+// CHECK_F: !llvm.linker.options = !{![[MODULE:[0-9]+]]}
+// CHECK_F: ![[MODULE]] = !{!"-framework", !"SomeKit"}
+#ifdef F
+@import OtherKit;
```
jrose-apple, line 43: What's the difference between E and F? Also, it looks like OtherKit is missing from the PR.

jrose-apple (commented):
This seems reasonable. I was worried at first about having a simple boolean in the module for this, since it seems like it would be easy to forget to update it, but maybe it's not an issue in practice. (My original idea was that there'd be a function that resolved it lazily, which you should call when asking about the LinkLibraries.)

lib/CodeGen/CodeGenModule.cpp
bcardosolopes: 'use' declarations can add more libraries, it seems to me those should still show up.

lib/Frontend/CompilerInstance.cpp
bcardosolopes: Yes. It's very common that we build a module and then read it back in the same compiler invocation. This would load the module again with the flag off (since we don't serialize it). This is also the reason I'm using names to keep track instead of 'Module *', since I wanna be sure we are always looking at the most recent load.

include/clang/Lex/ModuleMap.h
bcardosolopes: The reason I used a StringSet is in the ASTReader answer below.

lib/Lex/ModuleMap.cpp
bcardosolopes: Not a problem if we are using the `StringSet`, but makes sense if we use `TinyPtrVector<Module *>`. My first version of the patch was doing that, but it became unnecessary after.

test/Modules/use-exportas-for-link.m
bcardosolopes: Defining F triggers a new include inside OtherKit. The OtherKit tests is what exposes the need for the ASTReader to do the same. I indeed forget to include that in the test, this is the relevant snippet:

--- /dev/null
+++ b/test/Modules/Inputs/exportas-link/OtherKit.framework/Headers/OtherKit.h
@@ -0,0 +1,6 @@
+#import <SomeKitCore/SomeKitCore.h>
+
+#ifdef F
+#import <SomeKit/SomeKit.h>
+#endif
+

bcardosolopes:
> This seems reasonable. I was worried at first about having a simple boolean in the module for this, since it seems like it would be easy to forget to update it, but maybe it's not an issue in practice. (My original idea was that there'd be a function that resolved it lazily, which you should call when asking about the LinkLibraries.)

Yes, the function idea we discussed would be ideal, but at the point we check for things in Swift the only access we have is the `Module`, and from a `Module` we don't have access to the other modules to do the right computation. Like I said before, we could move all the logic to IR gen in both compilers, which would be cleaner to read, but then we get the maintenance tradeoff

lib/CodeGen/CodeGenModule.cpp
jrose-apple: I'm not sure I agree. If we ever extended this to non-frameworks, you'd definitely *not* want them to show up (because that would defeat the purpose).

bcardosolopes: What happens if SomeKitCore is declaring use of module X that SomeKit isn't declaring in its module map? We don't add X to show up as -framework X even if we are only importing SomeKit?

jrose-apple: That sounds correct to me. `export_as` means that anything in SomeKitCore is going to be visible in SomeKit, without exception, which means that SomeKit must itself declare all the same `use` flags to link successfully.

(If we allowed `export_as` in non-top-level modules this would get trickier.)

@DougGregor, what do you think?

bcardosolopes:
Updated the patch after the round of comments and added some missing bits.

lib/CodeGen/CodeGenModule.cpp
DougGregor: I agree with @jrose-apple that we should not restrict to this specific module name. The re-exporting module `SomeKit` should be responsible for re-declaring any `use` flags from the modules it re-exports, because this is the module that's ultimately responsible for the full interface and it has to stand alone: the module with `export_as` (`SomeKitCore` in this example) might not be around in some configurations, so we can't rely on gathering information from its module map.

DougGregor (commented):
This is simpler than I'd expected.

bcardosolopes:
Doug's feedback addressed

DougGregor (approved):
Looks great now, thank you!

jrose-apple (approved):

PR #179, opened 2018-04-13
CodeGenCXX: support PreserveMostCC in MS ABI

bob-wilson:
Microsoft has reserved 'U' for the PreserveMostCC which is used in the
swift runtime.  Add support for this.  This allows the swift runtime to
be built for Windows again.

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@329025 91177308-0d34-0410-b5e6-96231b3b80d8


bob-wilson:
This is the same as https://github.com/apple/swift-clang/pull/173 except for swift-4.2-branch (which is where it should have been merged to begin with).

PR #181, opened 2018-04-20
Tweak search path for API notes for private top-level modules

jrose-apple:
This is most important for frameworks: we want to look for PrivateHeaders/Bar.apinotes rather than Headers/Bar.apinotes and PrivateHeaders/Bar_private.apinotes. (This is especially important because 'Bar' is probably actually something like 'FooKit_Private' in practice.)

Doing this means keeping track of whether a particular module came from a "private" module map. Right now that's just being done by checking the name of the module map file; we could switch to something more principled in the future if it matters.

Note that the naming convention of 'FooKit_Private' for a private top-level module may cause certain API notes files to be loaded twice on case-insensitive filesystems: once as the regular API notes for 'FooKit_Private', and once as the *private* API notes for 'FooKit'. However, this shouldn't be a problem in practice (other than a bit of extra I/O) because (1) most declarations will only show up in one of the two modules anyway, being ignored in the other, and (2) the API notes we have are /mostly/ idempotent (except for how they record replacements).

(There is a Name field in an API notes file, but it's not currently validated.)

rdar://problem/39417625

### DIFF
--- a/lib/APINotes/APINotesManager.cpp
+++ b/lib/APINotes/APINotesManager.cpp
@@ -255,21 +255,41 @@ bool APINotesManager::loadCurrentModuleAPINotes(
     if (module->IsFramework) {
       // For frameworks, we search in the "Headers" or "PrivateHeaders"
       // subdirectory.
+      //
+      // Public modules:
+      // - Headers/Foo.apinotes
+      // - PrivateHeaders/Foo_Private.apinotes
+      // Private modules:
+      // - PrivateHeaders/Bar.apinotes (except that 'Bar' probably already has
+      //   the word "Private" in it in practice)
       llvm::SmallString<128> path;
       path += module->Directory->getName();
-      unsigned pathLen = path.size();
 
-      llvm::sys::path::append(path, "Headers");
-      if (auto apinotesDir = fileMgr.getDirectory(path))
-        tryAPINotes(apinotesDir, /*wantPublic=*/true);
+      if (!module->ModuleMapIsPrivate) {
+        unsigned pathLen = path.size();
 
-      path.resize(pathLen);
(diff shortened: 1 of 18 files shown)


jrose-apple:
@bcardosolopes In addition to regular review, I'm happy to take the parts that aren't specific to API notes and move them up to LLVM…but they have no uses there at the moment. Can you think of a good way to justify it?

DougGregor (approved):

lib/APINotes/APINotesManager.cpp
```diff
@@ -255,21 +255,41 @@ bool APINotesManager::loadCurrentModuleAPINotes(
...
+      // Public modules:
+      // - Foo.apinotes
+      // - Foo_Private.apinotes
+      // Private modules:
+      // - Bar.apinotes (except that 'Bar' probably already has the word
+      //   "Private" in it in practice)
       tryAPINotes(module->Directory, /*wantPublic=*/true);
-      tryAPINotes(module->Directory, /*wantPublic=*/false);
+      if (!module->ModuleMapIsPrivate)
+        tryAPINotes(module->Directory, /*wantPublic=*/false);
```
jrose-apple, line 292: @DougGregor: I wanted to ask specifically about this change. It seems *reasonable* to me but I could see it surprising people too.

DougGregor: I think it's reasonable to say that private modules don't have a separate "private private" set of API notes, yes.

bcardosolopes:
@jrose-apple the change abstracts away the direct check for names, which per se is a good change IMO.

bcardosolopes (approved):

jrose-apple:
Committed the Module parts as [r330452](https://reviews.llvm.org/rC330452).

PR #182, opened 2018-04-20
[4.2] Tweak search path for API notes for private top-level modules

jrose-apple:
- **Explanation**: The rules for finding an apinotes file for a particular module were not designed with top-level private modules in mind, leading to absurdities like "PrivateHeaders/FooKit_Private_private.apinotes". Tweak those rules so that a private top-level module has one canonical place to look for API notes, with no public/private split.
- **Scope**: Only affects private modules with API notes.
- **Issue**: rdar://problem/39417625
- **Risk**: Very low. There are almost none of these modules, and the whole point of the "private" part is that it is stripped out when a framework is actually shipped to customers.
- **Testing**: Added a compiler regression test.
- **Reviewed by**: @DougGregor, @bcardosolopes

### DIFF
--- a/include/clang/Basic/Module.h
+++ b/include/clang/Basic/Module.h
@@ -261,6 +261,10 @@ public:
   /// and headers from used modules.
   unsigned NoUndeclaredIncludes : 1;
 
+  /// \brief Whether this module came from a "private" module map, found next
+  /// to a regular (public) module map.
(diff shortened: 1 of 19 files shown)


jrose-apple:
@swift-ci Please test

jrose-apple:
@swift-ci Please nominate

PR #184, opened 2018-04-28
Add a new attribute swift_bridged_typedef for typedef bridging.

DougGregor:
When importing a typedef of a bridged type, Swift will use the
typedef-name to describe the unbridging version of the type (e.g.,
NSString). This attribute signals when Swift should instead import the
typedef using the bridged type (e.g., Swift's String).

Clang side of rdar://problem/39497900.


DougGregor:
@swift-ci please test and merge

PR #190, opened 2018-05-17
[APINotes] Warn when a private API notes file uses the wrong case

jrose-apple:
It's "FooKit_private.apinotes", not "FooKit_Private.apinotes".

rdar://problem/39914779


jrose-apple:
So this is a nice little change, though obviously not important…but the warning only comes up when the module is rebuilt. I guess that makes sense, but is there an easy way to get it played back later when the module is loaded? Like diagnostics in files? (Or does that not happen either, and I'm misremembering?)

DougGregor (approved):
This looks great. The limitation that warnings only show up at module build time, not when importing the module, has always been there for modules and precompiled headers. I wouldn't worry about it here.

PR #191, opened 2018-05-17
[4.2] [APINotes] Warn when a private API notes file uses the wrong case

jrose-apple:
- **Explanation**: Private API notes files for a module "FooKit" are found under the name "FooKit_private.apinotes". On a case-insensitive filesystem, this can be found under the name "FooKit_Private.apinotes"—a reasonable mistake! But one that causes problems on case-sensitive filesystems. Catch this common mistake with a warning.

- **Scope**: Only affects modules with private API notes.

- **Issue**: rdar://problem/39914779

- **Risk**: Low. This is a warning unrelated to pretty much everything else, and it only has a narrow scope anyway.

- **Testing**: Added a compiler regression test.


jrose-apple:
@swift-ci Please test

PR #192, opened 2018-05-18
Adjust for Clang r332598: VersionTuple always uses dots now

bob-wilson:
Remove references to the usesUnderscores() and UseDotAsSeparator()
functions that are no longer needed.


bob-wilson:
I'm going to go ahead and merge this to fix the build. Doug, can you check to make sure nothing more is needed here?

DougGregor:
This looks correct to me; we have never said it's "okay" to use underscores in API notes and I don't see any uses of this anywhere.

PR #194, opened 2018-05-23
ppc64le port.

asowani:
Support for ppc64le was added to stable branch, however same changes must go to "swift-4.1-branch" as well.


bob-wilson:
The Swift 4.1 release happened a while ago. We don't generally modify a release branch after the release is done. Can you just work on the 4.2 branch now?

asowani:
Sure. I will check if these changes are already in 4.2, else work on that branch. Thanks!

PR #195, opened 2018-06-12
[Refactor][Extract] The return type for functions should support block pointer types

hyp:
Instead of `static void (^)(int) extracted(ViewController *object) {` we now emit `static void (^extracted(ViewController *object))(int) {`. 

Also, local block return values must be copied when ARC is disabled. I went for a pessimistic copy that is not needed for non-local blocks, but I think it's fair to use it because we don't have the resources to create special heuristics for analyzing if a block is local or not for non-ARC code.

### DIFF
--- a/lib/Tooling/Refactor/Extract.cpp
+++ b/lib/Tooling/Refactor/Extract.cpp
@@ -1853,20 +1853,22 @@ llvm::Expected<RefactoringResult> ExtractOperation::perform(
       OS << "static ";
     else if (!isMethodExtraction())
       OS << (isInHeader(FunctionExtractionLoc, SM) ? "inline " : "static ");
-    ReturnType.print(OS, PP);
-    OS << ' ';
+    std::string QualifiedName;
+    llvm::raw_string_ostream NameOS(QualifiedName);
     if (isMethodExtraction() && IsDefinition)
-      printEnclosingMethodScope(FunctionLikeParentDecl, OS, PP);
-    unsigned NameOffset = OS.str().size();
-    OS << ExtractedName << '(';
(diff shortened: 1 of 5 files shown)


lib/Tooling/Refactor/Extract.cpp
```diff
@@ -1853,20 +1853,22 @@ llvm::Expected<RefactoringResult> ExtractOperation::perform(
       OS << "static ";
     else if (!isMethodExtraction())
       OS << (isInHeader(FunctionExtractionLoc, SM) ? "inline " : "static ");
-    ReturnType.print(OS, PP);
-    OS << ' ';
+    std::string QualifiedName;
+    llvm::raw_string_ostream NameOS(QualifiedName);
```
jkorous-apple, line 1857: Just an idea - wouldn't it be more transparent to actually construct the newly created function signature by using FunctionProtoType? My naive idea would be to construct a function type with argument types specified by CapturedVariables"."ParameterType, return type specified by ReturnType and qualifier optionally specified by FunctionLikeParentDecl.

https://clang.llvm.org/doxygen/classclang_1_1FunctionProtoType.html

hyp: It's possible but I don't want to do it in this repository. I can work on this while upstreaming the rest of this code.

vedantk (approved):
LGTM, although it might help to add a negative test for the `!IsExpr && !ARC && IsBlockPointer` case.

As a separate note not directly related to this review, I think it'd be helpful to split up ::perform() into smaller pieces. As-is, it takes a while to build up a roadmap of what the method should accomplish.

PR #199, opened 2018-06-21
[4.2] [APINotes] Don't apply API notes to @class / forward @protocol decls

jrose-apple:
- **Explanation**: Normally Swift ignores forward declarations of Objective-C classes, but SourceKit's module interface generation feature makes dummy ClassDecls for them instead. This exposes an issue where we allowed API notes to be placed on those forward declarations, even though the module may not own the types in question. This led to crashes in other parts of the Swift compiler that weren't expecting the attributes that API notes provide. This change sidesteps the whole issue by not allowing API notes to apply to forward declarations of Objective-C classes or protocols, consistent with a similar change made for structs, enums, and unions last year.

- **Scope**: Only affects classes and protocols with API notes, and even then only the forward declarations thereof.

- **Issue**: rdar://problem/40278479

- **Risk**: Low. This only affects SourceKit, and only for classes that didn't have definitions elsewhere where someone actually tried (perhaps accidentally) to use API notes anyway.

- **Testing**: Added compiler regression tests, verified that the SourceKit request works successfully. I also plan to add a Swift-side test on the master branch.

### DIFF
--- a/lib/Sema/SemaDeclObjC.cpp
+++ b/lib/Sema/SemaDeclObjC.cpp
@@ -1754,7 +1754,6 @@ Sema::ActOnForwardProtocolDeclaration(SourceLocation AtProtocolLoc,
       = ObjCProtocolDecl::Create(Context, CurContext, Ident, 
                                  IdentPair.second, AtProtocolLoc,
                                  PrevDecl);
-    ProcessAPINotes(PDecl);
 
     PushOnScopeChains(PDecl, TUScope);
     CheckObjCDeclScope(PDecl);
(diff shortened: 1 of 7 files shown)


jrose-apple:
@swift-ci Please test

jrose-apple:
Uh, hm. Right, we *are* using this in Swift, for terrible reasons.

DougGregor:
Oh, the `ObjectiveC` module depends on this, IIRC, because it forward-declares some Foundation classes.

jrose-apple:
apple/swift#17412
@swift-ci Please test macOS

jrose-apple:
Looks like the test-with-PR didn't actually work in this repo. I'll try it in the other direction.

jrose-apple:
Okay, the supporting commits should be merged in everywhere by now. Let's try this again.

@swift-ci Please test

PR #200, opened 2018-07-07
Fix a bad merge conflict resolution

bob-wilson:
The upstream-with-swift branch has refactored some code into a separate
function where the "method" parameter is named "M", but the conflict was
resolved without adjusting the name.


bob-wilson:
I merged this so that the code will at least compile, but Alex and Erik should review it. With Alex's change for rdar://32549953, the code that Erik modified is called from other locations. I don't know if those other locations should do the same check or a different one.

PR #209, opened 2018-09-17
IndexStore: repair Windows build

compnerd:
Building IndexStore would fail as `dllimport` would be applied to a non-inlined
function definition which is not permitted.  Remove the declaration to change
the error to a warning (inconsistent dll storage).  This allows us to build
correctly.  We could always apply the `dllexport` attribute to the definitions
to avoid the warning.

### DIFF
--- a/tools/IndexStore/IndexStore.cpp
+++ b/tools/IndexStore/IndexStore.cpp
@@ -530,7 +530,7 @@ indexstore_record_reader_search_symbols_f(indexstore_record_reader_t rdr,
   return reader->searchDecls(filterFn, receiverFn);
 }
 
-INDEXSTORE_PUBLIC bool
+bool
 indexstore_record_reader_symbols_apply_f(indexstore_record_reader_t rdr,
                                          bool nocache,
                                          void *context,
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @jrose-apple @akyrtzi

akyrtzi (changes requested):
Ah that was my bad, thanks for addressing this!
But I'm not a huge fan of the rest of the whitespace changes, could you just remove the 'INDEXSTORE_PUBLIC' identifiers so we can have a commit that only does one thing (remove 'INDEXSTORE_PUBLIC' to fix the windows build) ?

compnerd:
@akyrtzi - sure ... though the whitespace changes are clang-format :-)

akyrtzi (approved):

akyrtzi:
To clarify, with "I'm not a huge fan of the rest of the whitespace changes" I did not mean that I don't like the formatting changes themselves, what I don't like is adding unrelated whitespace changes to your fix.

compnerd:
@jrose-apple, @tkremenek - whats the best way to get this cherry-picked to stable?

bob-wilson:
Create a PR for the swift-5.0-branch to cherry-pick it

PR #210, opened 2018-09-18
IndexStore: repair Windows build

compnerd:
Building IndexStore would fail as `dllimport` would be applied to a non-inlined
function definition which is not permitted.  Remove the declaration to change
the error to a warning (inconsistent dll storage).  This allows us to build
correctly.  We could always apply the `dllexport` attribute to the definitions
to avoid the warning.


compnerd:
CC: @bob-wilson @akyrtzi

akyrtzi (approved):

PR #211, opened 2018-10-01
IndexStore: correct symbol visibility

compnerd:
On ELF targets, use protected visibility (symbols cannot be interpositioned).
On MachO, use default visibility.  On Windows, we need to define it to either
`__declspec(dllexport)` when building the library itself, or
`__declspec(dllimport)` when using the library.  This fixes a number of warnings
when building IndexCore for Windows.

### DIFF
--- a/include/indexstore/indexstore.h
+++ b/include/indexstore/indexstore.h
@@ -53,8 +53,12 @@
 #endif
 
 #ifndef INDEXSTORE_PUBLIC
-# if defined (_MSC_VER)
-#  define INDEXSTORE_PUBLIC __declspec(dllimport)
+# ifdef _WIN32
+#  ifdef IndexStore_EXPORTS
+#    define INDEXSTORE_PUBLIC __declspec(dllexport)
+#  else
+#    define INDEXSTORE_PUBLIC __declspec(dllimport)
+#  endif
 # else
 #  define INDEXSTORE_PUBLIC
 # endif
--- a/tools/IndexStore/CMakeLists.txt
+++ b/tools/IndexStore/CMakeLists.txt
@@ -12,9 +12,9 @@ set(LIBS
   clangIndexDataStore
 )
 
-set(LLVM_EXPORTED_SYMBOL_FILE ${CMAKE_CURRENT_SOURCE_DIR}/IndexStore.exports)
-
-set(ENABLE_SHARED SHARED)
+if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
+  set(LLVM_EXPORTED_SYMBOL_FILE ${CMAKE_CURRENT_SOURCE_DIR}/IndexStore.exports)
+endif()
 
 if(WIN32)
   set(output_name "libIndexStore")
@@ -22,7 +22,7 @@ else()
   set(output_name "IndexStore")
 endif()
 
-add_clang_library(IndexStore ${ENABLE_SHARED} ${ENABLE_STATIC}
+add_clang_library(IndexStore SHARED
   OUTPUT_NAME ${output_name}
   ${SOURCES}
 
@@ -35,31 +35,19 @@ add_clang_library(IndexStore ${ENABLE_SHARED} ${ENABLE_STATIC}
   Support
   )
 
-set(INDEXSTORE_LIBRARY_VERSION "${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}")
+if(APPLE)
+  set(INDEXSTORE_LIBRARY_VERSION "${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}")
 
-if(ENABLE_SHARED)
-  if(WIN32)
-    set_target_properties(IndexStore
-      PROPERTIES
-      VERSION ${INDEXSTORE_LIBRARY_VERSION}
-      DEFINE_SYMBOL _CINDEX_LIB_)
-  elseif(APPLE)
-    set(INDEXSTORE_LINK_FLAGS " -Wl,-compatibility_version -Wl,1")
-    set(INDEXSTORE_LINK_FLAGS "${INDEXSTORE_LINK_FLAGS} -Wl,-current_version -Wl,${INDEXSTORE_LIBRARY_VERSION}")
+  set(INDEXSTORE_LINK_FLAGS " -Wl,-compatibility_version -Wl,1")
+  set(INDEXSTORE_LINK_FLAGS "${INDEXSTORE_LINK_FLAGS} -Wl,-current_version -Wl,${INDEXSTORE_LIBRARY_VERSION}")
 
-    check_include_files("CoreServices/CoreServices.h" HAVE_CORESERVICES_H)
-    if(HAVE_CORESERVICES_H)
-      set(INDEXSTORE_LINK_FLAGS "${INDEXSTORE_LINK_FLAGS} -framework CoreServices")
-    endif()
-
-    set_property(TARGET IndexStore APPEND_STRING PROPERTY
-                 LINK_FLAGS ${INDEXSTORE_LINK_FLAGS})
-  else()
-    set_target_properties(IndexStore
-      PROPERTIES
-      VERSION ${INDEXSTORE_LIBRARY_VERSION}
-      DEFINE_SYMBOL _CINDEX_LIB_)
+  check_include_files("CoreServices/CoreServices.h" HAVE_CORESERVICES_H)
+  if(HAVE_CORESERVICES_H)
+    set(INDEXSTORE_LINK_FLAGS "${INDEXSTORE_LINK_FLAGS} -framework CoreServices")
   endif()
+
+  set_property(TARGET IndexStore APPEND_STRING PROPERTY
+               LINK_FLAGS ${INDEXSTORE_LINK_FLAGS})
 endif()
 
 if (LLVM_INSTALL_TOOLCHAIN_ONLY)


compnerd:
CC: @bitjammer @akyrtzi @jrose-apple

include/indexstore/indexstore.h
```diff
@@ -53,9 +53,13 @@
 
 #ifndef INDEXSTORE_PUBLIC
 # if defined (_MSC_VER)
```
akyrtzi: Nitpick, but slightly better to use `ifdef`.

include/indexstore/indexstore.h
```diff
@@ -53,9 +53,13 @@
...
-#  define INDEXSTORE_PUBLIC __declspec(dllimport)
+#  if defined(_CINDEX_LIB_)
```
akyrtzi: `_CINDEX_LIB_` came from mindlessly copy-pasting the CMake configuration from libclang.
This should be its own name, like `_CINDEXSTORE_LIB_`.

include/indexstore/indexstore.h
```diff
@@ -53,9 +53,13 @@
...
+#    define INDEXSTORE_PUBLIC __declspec(dllexport)
+#  else
+#    define INDEXSTORE_PUBLIC __declspec(dllimport)
+#  endif
 # else
-#  define INDEXSTORE_PUBLIC
+#  define INDEXSTORE_PUBLIC __attribute__((__visibility__("public")))
```
akyrtzi: Why is this change necessary ?

akyrtzi (changes requested):

include/indexstore/indexstore.h
compnerd: It's not strictly necessary, but does mean that we can actually remove the exports list (which is problematic for Windows, since we need to ensure that all symbols listed are defined, which is not the case when built without blocks).

include/indexstore/indexstore.h
compnerd: Ah, okay, I'll fix that.

include/indexstore/indexstore.h
akyrtzi: The export list is useful to ensure that only the listed symbols are exported and nothing else. Any other symbols would be prevented from being exported, making them private details of the library. I don't think it's a good idea to remove the export list.

compnerd: Right, with the suggested change, only the symbols marked with `INDEXSTORE_PUBLIC` would be exported and nothing else.  Okay,  I suppose that we can disable the exports list on Windows then.

akyrtzi: I thought for some reason that Windows doesn't use the export list (that it's unix specific) and we get things exported by marking them `__declspec(dllexport)` ?
If windows does use the export list I'd be fine with changing windows to not use it.

compnerd: Yeah, Windows does get it through the `__declspec(dllexport)` (on ELF and MachO targets, `__attribute__((__visibility__("default")))` with `-fvisibility-default=hidden` and `-fvisibility-inlines-hidden` will accomplish similar behaviour).  At that point, the source contains the actual source of truth.

compnerd:
@akyrtzi - okay, the current change will build everything with hidden visibility and still use the exports list on Linux and MachO (although, it will give the same results with and without it).  This also works for windows.  I fixed the `_CINDEX_LIB_` thing and simplified the CMake a bit since I removed the use of `_CINDEX_LIB_`.

include/indexstore/indexstore.h
akyrtzi: You would have to apply the new compiler options to _everything_, including llvm and clang, not just swift. And if then you'd have to convince everyone to mark explicitly the symbols that they want exported.
I don't think it is worth it, I generally like the greater control that the export list give us, there's a specific list of things to export and nothing else.

include/indexstore/indexstore.h
```diff
@@ -53,10 +53,14 @@
 #endif
 
 #ifndef INDEXSTORE_PUBLIC
-# if defined (_MSC_VER)
-#  define INDEXSTORE_PUBLIC __declspec(dllimport)
+# if defined(_WIN32)
```
akyrtzi: I'd suggest that a bit better is to use `# ifdef _WIN32` (same for `#  ifdef IndexStore_EXPORTS` below). `defined(MACRO)` is usually used when you want to check multiple macros in the same line.

include/indexstore/indexstore.h
```diff
@@ -53,10 +53,14 @@
...
+#  if defined(IndexStore_EXPORTS)
```
akyrtzi: `IndexStore_EXPORTS` doesn't seem that it is defined in the CMake file.

tools/IndexStore/CMakeLists.txt
```diff
@@ -34,32 +34,24 @@ add_clang_library(IndexStore ${ENABLE_SHARED} ${ENABLE_STATIC}
   Core
   Support
   )
+set_target_properties(IndexStore
+                      PROPERTIES
+                        CXX_VISIBILITY_PRESET hidden
+                        CXX_VISIBILITY_INLINES_HIDDEN YES)
```
akyrtzi: I think I wasn't clear in my previous comments. Setting these compiler arguments only for the files of `IndexStore` doesn't work because all the symbols coming from rest of llvm/clang/swift are still having 'default' visibility. These arguments are useful only if you can apply them across everything, so visibility is affected for every symbol.

If you want to see for yourself that this is not working for reducing the exported symbols, try this. Build `libIndexStore.dylib` on Darwin without using an exports list and then run this command on the dylib:

`nm -g /path/to/lib/libIndexStore.dylib | grep ' T '`

this command will show you what symbols get exported by the dylib. Without the export list you'll see a huge number of C++ symbols coming from llvm/clang/swift.

akyrtzi (changes requested):

include/indexstore/indexstore.h
compnerd: Sure, I can change that.  I agree that `defined()` is for when you want to check multiple macros, but that allows the original check to be left untouched :-)

include/indexstore/indexstore.h
compnerd: Right, it is defined implicitly by CMake (`add_library` adds a <NAME>_EXPORTS` macro).  See https://github.com/Kitware/CMake/blob/9fc3024f62304a65c2fb1eaf5303e417530c650d/Source/cmGeneratorTarget.cxx#L1733

tools/IndexStore/CMakeLists.txt
compnerd: Ah!  Right, the static libraries that we are linking will expose symbols too :-(.  Okay, I'll remove that.

include/indexstore/indexstore.h
akyrtzi: Interesting, I assume this is windows-only ? Such a macro is not defined for darwin.

include/indexstore/indexstore.h
```diff
@@ -53,10 +53,14 @@
...
-#  define INDEXSTORE_PUBLIC __declspec(dllimport)
+# ifdef _WIN32
+#  ifdef IndexStore_EXPORTS
+#    define INDEXSTORE_PUBLIC __declspec(dllexport)
+#  else
+#    define INDEXSTORE_PUBLIC __declspec(dllimport)
+#  endif
 # else
-#  define INDEXSTORE_PUBLIC
+#  define INDEXSTORE_PUBLIC __attribute__((__visibility__("default")))
```
akyrtzi: After you removed the visibility options this has become redundant, since it's the default visibility.
I'd recommend to leave it blank as `#  define INDEXSTORE_PUBLIC` for consistency with the rest of llvm/clang.

tools/IndexStore/CMakeLists.txt
```diff
@@ -12,17 +12,17 @@ set(LIBS
   clangIndexDataStore
 )
 
-set(LLVM_EXPORTED_SYMBOL_FILE ${CMAKE_CURRENT_SOURCE_DIR}/IndexStore.exports)
-
-set(ENABLE_SHARED SHARED)
+if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
+  set(LLVM_EXPORTED_SYMBOL_FILE ${CMAKE_CURRENT_SOURCE_DIR}/IndexStore.exports)
+endif()
```
akyrtzi, line 17: I know I mentioned that I'd be fine with not using the export list for windows, but it was not clear to me why you wanted to do that ? I mean what is the benefit with *not* setting the exports list for windows ?

akyrtzi (changes requested):

tools/IndexStore/CMakeLists.txt
compnerd: Because when you build without `-fblocks` you need to remove the methods guarded by `__feature(blocks)`.  I don't think that it is worth the extra work of trying to do something to merge multiple export lists (and we do not have a way to conditionalize the list).  We *could* duplicate the list and have two copies but they can get out of sync, so that is less than ideal.  Removing it on windows and relying on the dllstorage attributes seems the best approach.

include/indexstore/indexstore.h
compnerd: Sure.  Although, in ELF, it is possible to optimize this and use `__attribute__((__visibility__("default")))` when building and `__attribute__((__visibility__("protected")))` when using.

tools/IndexStore/CMakeLists.txt
akyrtzi: Ok, thanks for the explanation!

include/indexstore/indexstore.h
akyrtzi: No-one is bothering to do that across llvm/clang, so I have to assume that it's not that important, or if some client finds it important they could change how default visibility works within the client itself.

compnerd:
@akyrtzi does this look good now?

akyrtzi (approved):

PR #212, opened 2018-10-10
IndexStore: correct symbol visibility

compnerd:
On ELF targets, use protected visibility (symbols cannot be interpositioned).
On MachO, use default visibility.  On Windows, we need to define it to either
`__declspec(dllexport)` when building the library itself, or
`__declspec(dllimport)` when using the library.  This fixes a number of warnings
when building IndexCore for Windows.

Remove the use of the exports list and rely on the annotations instead.  This
makes the build similar across all the targets (Windows requires all listed
symbols to be present, and we would need to preprocess the list otherwise).

Fix the accidental usage of `_CINDEX_LIB_` opting for the CMake standard
variable `<Name>_EXPORTS`.

Simplify the logic in the CMake for Apple targets a bit.

### DIFF
--- a/include/indexstore/indexstore.h
+++ b/include/indexstore/indexstore.h
@@ -53,8 +53,12 @@
 #endif
 
 #ifndef INDEXSTORE_PUBLIC
-# if defined (_MSC_VER)
-#  define INDEXSTORE_PUBLIC __declspec(dllimport)
+# ifdef _WIN32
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @akyrtzi @bob-wilson 

This is a cherry-pick into swift-5.0 to fix the Windows build.

akyrtzi (approved):

PR #213, opened 2018-10-25
CFRuntime ABIs (for CFString)

compnerd:
This cherry-picks a series of patches to add the ability to alter the CFObject layout for CFStrings.  This is needed for the CFString work that @millenomi is working on for swift-corelibs-foundation.


compnerd:
CC: @millenomi @gottesmm @bob-wilson

lib/CodeGen/CodeGenModule.cpp
```diff
@@ -4030,39 +4030,76 @@ CodeGenModule::GetAddrOfConstantCFString(const StringLiteral *Literal) {
...
-        GV->setLinkage(llvm::GlobalValue::ExternalLinkage);
-      } else {
-        GV->setDLLStorageClass(llvm::GlobalValue::DLLExportStorageClass);
-        GV->setLinkage(llvm::GlobalValue::ExternalLinkage);
-      }
+
+    switch (CFRuntime) {
+    default: break;
+    case LangOptions::CoreFoundationABI::Swift: LLVM_FALLTHROUGH;
+    case LangOptions::CoreFoundationABI::Swift5_0: LLVM_FALLTHROUGH;
```
millenomi: This is not correct; the class name for Swift 5 is $S… (uppercase).

lib/CodeGen/CodeGenModule.cpp
```diff
@@ -4094,18 +4136,21 @@ CodeGenModule::GetAddrOfConstantCFString(const StringLiteral *Literal) {
...
   // LTO.  Doing that changes the section it ends in, which surprises ld64.
-  if (getTriple().isOSBinFormatMachO())
+  if (Triple.isOSBinFormatMachO())
     GV->setSection(isUTF16 ? "__TEXT,__ustring"
                            : "__TEXT,__cstring,cstring_literals");
-
+  // Make sure the literal ends up in .rodata to allow for safe ICF and for
+  // the static linker to adjust permissions to read-only later on.
+  else if (Triple.isOSBinFormatELF())
+    GV->setSection(".rodata");
```
millenomi, line 4157: This is also incorrect. For Swift 5, the backing store of a CFString must be writable because we don’t have immortal objects yet.

millenomi (changes requested):

lib/CodeGen/CodeGenModule.cpp
millenomi: Ah, I misread. If it’s for the C string literal, then it’s okay.

compnerd: Yeah, this is for the literal data.  The data itself is mutable (which is also the case for the CFString objects with the ObjC interop (there is, in fact, a test that tests exactly that :-)).

lib/CodeGen/CodeGenModule.cpp
compnerd: Oh, bleh.  I'll fix that upstream, and fix that here too.

lib/CodeGen/CodeGenModule.cpp
```diff
@@ -4073,7 +4110,12 @@ CodeGenModule::GetAddrOfConstantCFString(const StringLiteral *Literal) {
   Fields.add(cast<llvm::ConstantExpr>(CFConstantStringClassRef));
 
   // Flags.
-  Fields.addInt(IntTy, isUTF16 ? 0x07d0 : 0x07C8);
+  if (IsSwiftABI) {
+    Fields.addInt(IntPtrTy, IsSwift4_1 ? 0x05 : 0x01);
+    Fields.addInt(Int64Ty, isUTF16 ? 0x07d0 : 0x07c8);
```
millenomi, line 4120: This will work on the platforms we ship on swift.org, but may not on big endian? But that can be a follow-up.

lib/CodeGen/CodeGenModule.cpp
compnerd: SVN r345291

millenomi (approved):

millenomi:
this satisfies what SCF needs, but cc @gottesmm @bob-wilson for a second compiler set of eyes on this.

bob-wilson:
I'm not the best person to review these cherry-picks. Duncan is the release manager for swift-lang, so I added him as a reviewer.

millenomi:
@dexonsmith Do let us know?

millenomi:
(Whoops, he may be out of office?)

dexonsmith:
> @dexonsmith Do let us know?

Sorry for the delay—@ahatanaka says he can take a look this afternoon.

gottesmm:
Thanks @dexonsmith!

millenomi:
@ahatanaka please ping me once you've looked at this?

lib/AST/ASTContext.cpp
```diff
@@ -5761,50 +5761,86 @@ int ASTContext::getIntegerTypeOrder(QualType LHS, QualType RHS) const {
...
+  struct {
+    QualType Type;
+    const char *Name;
+  } Fields[5];
+  unsigned Count = 0;
+
+  /// Objective-C ABI
+  ///
+  ///    typedef struct __NSConstantString_tag {
+  ///      const char *isa;
```
ahatanaka: Since IntTy.withConst() is called below, should this be "const int *isa"?

ahatanaka (approved):
I have one minor comment, but patch looks good to me.

lib/AST/ASTContext.cpp
compnerd: Yes, it should be a `const int *` in the comment.  Good eye!  I'll fix that :-).

PR #214, opened 2018-10-29
AST: extend MS decoration handling for extended vectors

compnerd:
We correctly handled extended vectors of non-floating point types.
However, we have the Intel style builtins which MSVC also supports which
do overlap in sizes with the floating point extended vectors.  This
would result in overloading of floating point extended vector types
which matched sizes (e.g. <3 x float> would be backed by a <4 x float>
and thus match sizes) to be mangled similarly.  Extended vectors are a
clang extension which live outside of the builtins, so mangle them all
similarly.  This change just extends the current scheme to treat
floating point types similar to the way that we treat other types
currently.

This now allows the swift runtime to be built for Windows again.

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@345479 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 92a8170754dda7ca8769e784388396cc267fc68b)

### DIFF
--- a/lib/AST/MicrosoftMangle.cpp
+++ b/lib/AST/MicrosoftMangle.cpp
@@ -2502,18 +2502,20 @@ void MicrosoftCXXNameMangler::mangleType(const VectorType *T, Qualifiers Quals,
   // Pattern match exactly the typedefs in our intrinsic headers.  Anything that
   // doesn't match the Intel types uses a custom mangling below.
   size_t OutSizeBefore = Out.tell();
-  llvm::Triple::ArchType AT =
-      getASTContext().getTargetInfo().getTriple().getArch();
-  if (AT == llvm::Triple::x86 || AT == llvm::Triple::x86_64) {
-    if (Width == 64 && ET->getKind() == BuiltinType::LongLong) {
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @bob-wilson @DougGregor 

This fixes clang's handling for extended vector types with MS ABI.  This gets triggered when compiling with template specializations involving `__attribute__((__ext_vector__(\d)))`.

compnerd:
CC: @tkremenek

tkremenek:
Just a heads up that `swift-5.0-branch` will re-branch from `master` on November 16.  We can explicitly pull changes like this into `swift-5.0-branch` now, or just wait for the re-branch from `master`.

tkremenek:
@swift-ci test

tkremenek:
@dexonsmith this looks reasonable to me.  WDYT?

tkremenek:
Build failure looks unrelated.

tkremenek:
@swift-ci clean test

dexonsmith:
LGTM.

dexonsmith (approved):

PR #218, opened 2018-10-31
Adjust for SVN r345637

compnerd:
ObjC1 and ObjC2 were removed as ObjC1 has not been supported in a very
long time.  It has been replaced with `ObjC`.


compnerd:
CC: @bob-wilson

compnerd:
This probably is further out than stable-next, whats the right branch @bob-wilson?

bob-wilson:
We don't use stable-next. This should go to upstream-with-swift

compnerd:
@bob-wilson - rebased :-)

PR #230, opened 2018-11-04
Swift 5.0 branch fix build

gottesmm:
Fixing the build.


gottesmm:
@swift-ci test

gottesmm:
The build-script line on macOS is:

/Users/buildnode/jenkins/workspace/swift-clang-PR-osx/swift/utils/build-script --test --validation-test --lit-args=-v --compiler-vendor=apple --reconfigure --verbose-build --build-ninja --build-swift-stdlib-unittest-extra --ios --tvos --watchos --lldb --lldb-no-debugserver --lldb-use-system-debugserver --lldb-build-type=Release --lldb-test-swift-only --lldb-assertions --build-subdir=buildbot_incremental --release --assertions --llbuild --swiftpm --swiftsyntax --playgroundsupport --sil-verify-all --skip-test-ios-host --skip-test-tvos-host --skip-test-watchos-host.

So we will compile enough to expose the problem if I did not revert enough. But I think these two are the only problems.

gottesmm:
@swift-ci test linux platform

gottesmm:
Did some more spelunking. It looks like the reason why linux is failing is we are testing on the 5.0 branch... the 5.0 branch on Linux is already borked for other reasons.

When I look at the swift-5.0 branch:

https://ci.swift.org/view/swift-5.0-branch/job/oss-swift-5.0-incremental-RA-osx/

There is a really clear blame list that this is breaking the build. See build #35 where these are the only two commits. So this should unbreak the swift-5.0-branch os x build. Since linux is already failing, we shouldn't hold up fixing macOS for that.

PR #235, opened 2018-11-20
[Swift+WASM] Adopt SwiftABIInfo for WebAssembly.

ddunbar:

PR #236, opened 2018-11-20
[Swift+WASM] Add an ABI hook for whether swiftcc is supported.

ddunbar:
- This adds an extension point so the ABI can declare whether or not it supports the swift calling convention. Open to suggestions if you think this doesn't belong here (or is flat out the wrong thing to try and to do bring up wasm).

 - [ ] Depends on #235.


DougGregor:
I feel like @rjmccall is in a better position to review this.

rjmccall:
I've resisted doing this on other targets because:
- it's basically trivial to support the CC on a target by just using the standard rules for your platform (or, in this case, ignore it when preparing wasm output);
- it's not at all unreasonable to think that a backend might want to give special treatment to swifterror and swiftself, and doing so is much easier if they only appear in conjunction with swiftcc; and 
- it's much easier to write portable tests for frontend IR output if as much as possible is consistent across targets.

ddunbar:
@rjmccall that makes a lot of sense, thanks. Will investigate that angle instead.

PR #237, opened 2018-11-24
clang-cl: Add "/clang:" pass-through arg support.

compnerd:
The clang-cl driver disables access to command line options outside of the
"Core" and "CLOption" sets of command line arguments. This filtering makes it
impossible to pass arguments that are interpreted by the clang driver and not
by either 'cc1' (the frontend) or one of the other tools invoked by the driver.

An example driver-level flag is the '-fno-slp-vectorize' flag, which is
processed by the driver in Clang::ConstructJob and used to set the cc1 flag
"-vectorize-slp". There is no negative cc1 flag or -mllvm flag, so it is not
currently possible to disable the SLP vectorizer from the clang-cl driver.

This change introduces the "/clang:" argument that is available when the
driver mode is set to CL compatibility. This option works similarly to the
"-Xclang" option, except that the option values are processed by the clang
driver rather than by 'cc1'. An example usage is:

  clang-cl /clang:-fno-slp-vectorize /O2 test.c

Another example shows how "/clang:" can be used to pass a flag where there is
a conflict between a clang-cl compat option and an overlapping clang driver
option:

  clang-cl /MD /clang:-MD /clang:-MF /clang:test_dep_file.dep test.c

In the previous example, the unprefixed /MD selects the DLL version of the msvc
CRT, while the prefixed -MD flag and the -MF flags are used to create a make
dependency file for included headers.

One note about flag ordering: the /clang: flags are concatenated to the end of
the argument list, so in cases where the last flag wins, the /clang: flags
will be chosen regardless of their order relative to other flags on the driver
command line.

Patch by Neeraj K. Singh!

Differential revision: https://reviews.llvm.org/D53457

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@346393 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 8bf07130276f736f15ff2bd346c604a17cd6cc85)

### DIFF
--- a/docs/UsersManual.rst
+++ b/docs/UsersManual.rst
@@ -2776,6 +2776,7 @@ Execute ``clang-cl /?`` to see a list of supported options:
       /arch:<value>           Set architecture for code generation
       /Brepro-                Emit an object file which cannot be reproduced over time
       /Brepro                 Emit an object file which can be reproduced over time
+      /clang:<arg>            Pass <arg> to the clang driver
       /C                      Don't discard comments when preprocessing
       /c                      Compile only
       /d1PP                   Retain macro definitions in /E mode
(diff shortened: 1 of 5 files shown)


compnerd:
CC: @dexonsmith - nominating this change for the 5.0 branch.  This is needed to repair some of the compilation issues for Windows.  It is a very low risk change.  The change itself has been upstream for some time.  It simply adds a new command line option to the cl driver in clang.  The normal compiler code path is not touched.

compnerd:
CC: @tkremenek

hyp (approved):
LGTM

dexonsmith (approved):

PR #240, opened 2018-11-28
Revert "Edit: adjust for SVN r347417"

bob-wilson:
This reverts commit 5159071da70265031a53fb491881edd9dd97b14d.

Clang r347417 was reverted, so this also needs to be reverted for now.


bob-wilson:
@compnerd FYI. We'll need to reapply this when/if the Clang change goes back in.

PR #241, opened 2018-11-29
Edit: adjust for SVN r347756 (was previously r347417)

bob-wilson:
The change in r347417 was reverted but then reapplied in r347756. Reapply the adjustment needed to go with that change.


bob-wilson:
@compnerd FYI

PR #242, opened 2018-12-07
[5.0] Only load private API notes if there are private submodules

jrose-apple:
- **Explanation**: We're moving away from a convention of private submodules (MyKit.Private) to private top-level modules (MyKit_Private). Unfortunately, the name of a private API notes file (MyKit_private.apinotes) and an API notes files for a private top-level module (MyKit_Private.apinotes) are the same on a case-insensitive filesystem. Disambiguate by checking if MyKit has any private submodules.

- **Scope**: Affects frameworks (and non-framework libraries) with private top-level modules. Does not affect frameworks with private submodules, or no private module map at all.

- **Issue**: rdar://problem/45705724&46311788

- **Risk**: Low. Both the old and new configurations are accounted for here, and if anyone's mixing them we'll at least do something consistent.

- **Testing**: Passed regression tests.

### DIFF
--- a/lib/APINotes/APINotesManager.cpp
+++ b/lib/APINotes/APINotesManager.cpp
@@ -245,6 +245,14 @@ static void checkPrivateAPINotesName(DiagnosticsEngine &diags,
   diags.Report(SourceLocation(), diagID) << module->Name << realFilename;
 }
 
+/// \returns true if any of \p module's immediate submodules are defined in a
+/// private module map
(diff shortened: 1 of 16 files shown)


jrose-apple:
@swift-ci Please test

dexonsmith (approved):
This LGTM.

bcardosolopes:
LGTM

bcardosolopes (approved):

PR #243, opened 2018-12-07
[5.0] [APINotes] `RetainCountConvention: none` overrules `arc_cf_audited`

jrose-apple:
- **Explanation**: The `RetainCountConvention: none` API note wasn't taking effect within `pragma clang arc_cf_code_audited begin`/`end` regions, because the pragma was applied separately. Counteract that by adding an opt-out attribute when `RetainCountConvention: none` is used.

- **Scope**: Affects this particular API note only, and the added attribute only makes a difference in `pragma clang arc_cf_code_audited begin`/`end` regions.

- **Issue**: rdar://problem/39619530

- **Risk**: Low

- **Testing**: Passed regression tests.


jrose-apple:
@swift-ci Please test

dexonsmith (approved):
LGTM.

bob-wilson:
We should never modify the stable branch directly. This needs to go to swift-5.0-branch

PR #247, opened 2018-12-17
[analyzer] ObjCDealloc: Fix a crash when a class attempts to deallocate a class.

haoNoQ:
@devincoughlin: review request for CCC.

rdar://problem/46376777

---

The checker wasn't prepared to see the dealloc message sent to the class itself
rather than to an instance, as if it was +dealloc.

Additionally, it wasn't prepared for pure-unknown or undefined self values.
The new guard covers that as well, but it is annoying to test because
both kinds of values shouldn't really appear and we generally want to
get rid of all of them (by modeling unknown values with symbols and
by warning on use of undefined values before they are used).

The CHECK: directive for FileCheck at the end of the test looks useless,
so i removed it.

Differential Revision: https://reviews.llvm.org/D55680

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@349228 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 2b29b6937bcb5f43b5572b70837a6374b43927c5)


devincoughlin:
LGTM. Thanks for fixing this!

PR #248, opened 2018-12-18
[analyzer] MoveChecker: Enable by default as cplusplus.Move.

haoNoQ:
@devincoughlin: review request for CCC.

rdar://problem/41349073


devincoughlin (approved):
This is quite large, but it looks good to me.

PR #249, opened 2018-12-19
Disable using `-fsanitize=leak` from the Clang Driver.

ghost:
LeakSanitizer (LSan) is not yet mature for Darwin so until this
changes we should avoid supporting it in Swift Clang and AppleClang.

rdar://problem/45841334

### DIFF
--- a/lib/Driver/ToolChains/Darwin.cpp
+++ b/lib/Driver/ToolChains/Darwin.cpp
@@ -2358,6 +2358,9 @@ SanitizerMask Darwin::getSupportedSanitizers() const {
   Res |= SanitizerKind::FuzzerNoLink;
   Res |= SanitizerKind::Function;
 
+  // Apple-Clang: Don't support LSan. rdar://problem/45841334
+  Res &= ~SanitizerKind::Leak;
+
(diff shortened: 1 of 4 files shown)


kubamracek (approved):

hyp (approved):

jrose-apple (dismissed):
This branch is used to build for Linux too. I think we even run Leak Sanitizer on Swift stuff there.

jrose-apple:
In general, `upstream-with-swift` should not have anything in it that isn't specific to Swift. Other Apple-specific stuff should stay Apple-internal at this time.

danliew-apple:
@jrose-apple This change is only for Darwin. Linux should remain unchanged by this PR.

danliew-apple:
@jrose-apple 
> In general, upstream-with-swift should not have anything in it that isn't specific to Swift. Other Apple-specific stuff should stay Apple-internal at this time.

This patch was originally intended to be internal but @hyp requested it be moved here.

delcypher:
@jrose-apple The commits shouldn't mention `Apple-internal` anymore. So I'll land this.

PR #250, opened 2018-12-20
[APINotes] Test that 'ResultType' preserves the 'instancetype' flag

jrose-apple:
It probably shouldn't, but we now have Apple frameworks relying on this behavior, and since `instancetype` isn't a real type you can't specify it using `ResultType` anyway. In practice, if we had to be locked into some behavior for `instancetype`-returning methods, this is probably the best one.


DougGregor (approved):
Sigh, yeah. It makes sense to codify this behavior now that we depend on it.

PR #252, opened 2019-01-05
CodeGen: fix autolink emission on ELF

compnerd:
The autolinking extension for ELF uses a slightly different format for
encoding the autolink information compared to COFF and MachO.  Account
for this in the CGM to ensure that we do not assert when emitting
assembly or an object file.

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@350476 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 4159746b2cb2077a5a10b1ba65d8c865d35f6bd3)

### DIFF
--- a/lib/CodeGen/CodeGenModule.cpp
+++ b/lib/CodeGen/CodeGenModule.cpp
@@ -1654,6 +1654,8 @@ static void addLinkOptionsPostorder(CodeGenModule &CGM, Module *Mod,
   // Add linker options to link against the libraries/frameworks
   // described by this module.
   llvm::LLVMContext &Context = CGM.getLLVMContext();
+  bool IsELF = CGM.getTarget().getTriple().isOSBinFormatELF();
+  bool IsPS4 = CGM.getTarget().getTriple().isPS4();
 
   // For modules that use export_as for linking, use that module
   // name instead.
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @akyrtzi @tkremenek 

This fixes an internal error when using modules with ELF and autolinking.  This is fairly important for s-p-m which enables modules when compiling, and causes failures on Linux.

compnerd:
CC: @bob-wilson

akyrtzi:
We are not going to put the clang executable into the swift-5.0 linux toolchain, so what is the benefit of taking the fix in the `swift-5.0-branch` ?

EDIT: To clarify a bit more, this issue does not appear to affect compiling swift source files, it seems to only affect the clang compiler using module maps with auto-link entries.

compnerd:
@akyrtzi - well, there was some discussion on the forums about that behaviour.  It seems that there generally was interest in the llvm toolchain being bundled with it.  In fact, the windows nightlies currently do bundle everything together.  I think that people are trying to use this build of clang to build packages with s-p-m and that does cause the backend to error out, so its just an experience thing.

akyrtzi:
\cc @dexonsmith for FYI. I tested clang and clang-tools-extra on Darwin with this change and all tests pass. I think it's safe and goodness to take.

akyrtzi:
@compnerd please also cherry-pick to `swift-5.1-branch`

PR #253, opened 2019-01-10
CodeGen: fix autolink emission on ELF

compnerd:
The autolinking extension for ELF uses a slightly different format for
encoding the autolink information compared to COFF and MachO.  Account
for this in the CGM to ensure that we do not assert when emitting
assembly or an object file.

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@350476 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 4159746b2cb2077a5a10b1ba65d8c865d35f6bd3)


compnerd:
CC: @akyrtzi @dexonsmith @tkremenek @bob-wilson

PR #254, opened 2019-01-16
[docs] Document the API Notes feature added for Swift

jrose-apple:
Hey, I finally did this!

rdar://problem/38123494

### DIFF
--- a/docs/APINotes.rst
+++ b/docs/APINotes.rst
@@ -1,0 +1,361 @@
+================================================
+API Notes: Annotations Without Modifying Headers
+================================================
+
+**The Problem:** You have headers you want to use, but you also want to add
+extra information to some of the APIs. You don't want to put that information
+in the headers themselves---perhaps because you want to keep them clean for
+other clients, or perhaps because they're from some open source project and you
+don't want to modify them at all.
+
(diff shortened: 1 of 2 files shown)


docs/APINotes.rst
```diff
@@ -0,0 +1,353 @@
...
+    - Selector: superclass
+      MethodKind: Class
+      NullabilityOfRet: O
+
+:Type:
+
+  Used for properties and globals. This completely overrides the type of the
+  declaration; it should ideally only be used for Swift backwards
+  compatibility, when existing type information has been made more precise in a
+  header. Prefer 'Nullability' and other annotations when possible.
```
DougGregor, line 231: I think it's worth a note about this really being the bare type, skipping any inferred attributes. For example, even iwithin an `NS_ASSUME_NONNULL_BEGIN`  / `NS_ASSUME_NONNULL_END` block, you would need to write `id _Nonnull`.

DougGregor (approved):
Wow, thank you!

docs/APINotes.rst
jrose-apple: Good point, will update.

dexonsmith (approved):

PR #261, opened 2019-02-22
CodeGen: use COMDAT for block copy/destroy helpers

compnerd:
SVN r339438 added support to deduplicate the helpers by using a consistent
naming scheme and using LinkOnceODR semantics.  This works on ELF by means of
weak linking semantics, and entirely does not work on PE/COFF where you end up
with multiply defined strong symbols, which is a strong error on PE/COFF.
Assign the functions a COMDAT group so that they can be uniqued by the linker.
This fixes the use of blocks in CoreFoundation on Windows.

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@354678 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 8f37e96206e47808d516401da00e100a960f66aa)

### DIFF
--- a/lib/CodeGen/CGBlocks.cpp
+++ b/lib/CodeGen/CGBlocks.cpp
@@ -2029,6 +2029,8 @@ CodeGenFunction::GenerateCopyHelperFunction(const CGBlockInfo &blockInfo) {
   llvm::Function *Fn =
     llvm::Function::Create(LTy, llvm::GlobalValue::LinkOnceODRLinkage,
                            FuncName, &CGM.getModule());
+  if (CGM.supportsCOMDAT())
+    Fn->setComdat(CGM.getModule().getOrInsertComdat(FuncName));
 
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @bob-wilson @ahatanaka @dexonsmith

compnerd:
This is needed to build Foundation for Windows

dexonsmith:
The commit LGTM, but I'm not sure if this is the correct branch to target.  I think we usually pull into the appropriate "swift-X.Y-branch" and rely on the automerger to land it in "stable".

compnerd:
@dexonsmith - ugh, right!  I'll rebase and change this to 5.1.  Sorry, I keep getting this screwed up in my head.

compnerd:
@dexonsmith - given the LGTM, going to merge this (its the same cherry-pick, moved to the 5.1 branch).

PR #262, opened 2019-02-28
Fix Linux build of DirectoryWatcher

benlangmuir:
Use the correct namespace.

rdar://48452986


benlangmuir:
@swift-ci please smoke test

PR #264, opened 2019-02-28
DirectoryWatcher: correct namespace

compnerd:
`RetryAfterSignal` is in the `llvm::sys` namespace, not the `llvm` namespace.
Add the subnamespace to fix the build.


compnerd:
CC: @jkorous-apple @bob-wilson @dexonsmith

PR #278, opened 2019-03-19
[Swift] Add support for __attribute__((objc_class_stub))

slavapestov:
A class annotated with this attribute requires the compiler to emit
a call to `objc_loadClassref()` when the class is referenced.

Categories are not affected.

rdar://36635199

### DIFF
--- a/lib/CodeGen/CGObjCMac.cpp
+++ b/lib/CodeGen/CGObjCMac.cpp
@@ -726,6 +726,28 @@ public:
                                      "objc_begin_catch");
   }
 
+  /// Class objc_loadClassref (void *)
+  ///
+  /// Loads from a classref. For Objective-C stub classes, this invokes the
+  /// initialization callback stored inside the stub. For all other classes
+  /// this simply dereferences the pointer.
+  llvm::Constant *getLoadClassrefFn() const {
+    // Add the non-lazy-bind attribute, since objc_loadClassref is likely to
+    // be called a lot.
+    //
+    // Also it is safe to make it readnone, since we never load or store the
+    // classref except by calling this function.
+    llvm::Type *params[] = { Int8PtrPtrTy };
+    return CGM.CreateRuntimeFunction(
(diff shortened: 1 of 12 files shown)


slavapestov:
The corresponding Swift change is https://github.com/apple/swift/pull/23320

lib/CodeGen/CGObjCMac.cpp
```diff
@@ -7247,19 +7266,38 @@ CGObjCNonFragileABIMac::EmitClassRefFromId(CodeGenFunction &CGF,
...
+        ClassGV = llvm::ConstantExpr::getAdd(ClassGV, Tag);
+        ClassGV = llvm::ConstantExpr::getIntToPtr(ClassGV,
+                                                  ObjCTypes.Int8PtrTy);
+      }
     } else {
       ClassGV = GetClassGlobal((getClassSymbolPrefix() + II->getName()).str(),
                                NotForDefinition);
     }
 
-    Entry = new llvm::GlobalVariable(CGM.getModule(), ObjCTypes.ClassnfABIPtrTy,
```
jrose-apple, line 7255: What happens to this type? Can we assert that it's the correct type, at least?

slavapestov: It's going to be Int8PtrTy in one case and ClassnfABIPtrTy in the other. I could pretend its a ClassnfABIPtrTy in both, but in the stub case the pointed-to-thing is not a Class, and the pointer is offset by 1.

jrose-apple: Yes, I see that, but in the old code we checked that we had a ClassnfABIPtrTy. Is it important to preserve that check in the normal case?

slavapestov: I added the assert. The Clang CL for this patch is here, you can post review comments there instead: https://reviews.llvm.org/D59628

PR #256, opened 2019-01-20
Mark indexstore callbacks with noescape attribute

kastiglione:
Annotate block and function pointer parameters in `indexstore.h` to be `noescape`. When wrapping the indexstore API with Swift, this removes the need for `@escaping` or `withoutActuallyEscaping()`.

PR #288, opened 2019-03-30
Mark indexstore enum flag types with flag_enum attribute

kastiglione:
Adds useful compiler attributes to the two indexstore bit flag types: `indexstore_symbol_property_t` and `indexstore_symbol_role_t`.

These `enum` declarations are now declared with the following features:

1. Defined as 64bit ints using C++ or ObjC sized enums, otherwise falling back to a `typedef`
2. Marked with `flag_enum` attribute, if available
3. Marked with `enum_extensibility(open)` attribute, if available

Additionally, the functions that return these types now return the `enum` itself or the fallback `typedef`. Previously these functions returned an explicit `uint64_t`.

The motivation for these changes is for use from Swift. Now, these bit flags are imported into Swift as [`OptionSet`](https://developer.apple.com/documentation/swift/optionset)s.

### DIFF
--- a/include/indexstore/indexstore.h
+++ b/include/indexstore/indexstore.h
@@ -85,9 +85,25 @@
 #endif
 
 #if __has_attribute(flag_enum)
-# define INDEXSTORE_OPTIONS __attribute__((flag_enum))
+# define _INDEXSTORE_FLAG_ENUM __attribute__((flag_enum))
 #else
-# define INDEXSTORE_OPTIONS
+# define _INDEXSTORE_FLAG_ENUM
+#endif
+
+#if __has_attribute(enum_extensibility)
+# define _INDEXSTORE_OPEN_ENUM __attribute__((enum_extensibility(open)))
+#else
+# define _INDEXSTORE_OPEN_ENUM
+#endif
+
+#define _INDEXSTORE_OPTIONS_ATTRS _INDEXSTORE_OPEN_ENUM _INDEXSTORE_FLAG_ENUM
+
+#if __has_extension(cxx_strong_enums)
+# define INDEXSTORE_OPTIONS(_name, _type) enum _INDEXSTORE_OPTIONS_ATTRS _name : _type
+#elif __has_feature(objc_fixed_enum)
+# define INDEXSTORE_OPTIONS(_name, _type) typedef enum _INDEXSTORE_OPTIONS_ATTRS _name : _type _name; enum _INDEXSTORE_OPTIONS_ATTRS _name : _type
+#else
+# define INDEXSTORE_OPTIONS(_name, _type) typedef _type _name; enum _INDEXSTORE_OPTIONS_ATTRS
 #endif
 
 INDEXSTORE_BEGIN_DECLS
@@ -269,7 +285,7 @@ typedef enum {
   INDEXSTORE_SYMBOL_SUBKIND_SWIFTACCESSORMODIFY = 1015,
 } indexstore_symbol_subkind_t;
 
-typedef enum INDEXSTORE_OPTIONS {
+INDEXSTORE_OPTIONS(indexstore_symbol_property_t, uint64_t) {
   INDEXSTORE_SYMBOL_PROPERTY_GENERIC                          = 1 << 0,
   INDEXSTORE_SYMBOL_PROPERTY_TEMPLATE_PARTIAL_SPECIALIZATION  = 1 << 1,
   INDEXSTORE_SYMBOL_PROPERTY_TEMPLATE_SPECIALIZATION          = 1 << 2,
@@ -279,7 +295,7 @@ typedef enum INDEXSTORE_OPTIONS {
   INDEXSTORE_SYMBOL_PROPERTY_GKINSPECTABLE                    = 1 << 6,
   INDEXSTORE_SYMBOL_PROPERTY_LOCAL                            = 1 << 7,
   INDEXSTORE_SYMBOL_PROPERTY_PROTOCOL_INTERFACE               = 1 << 8,
-} indexstore_symbol_property_t;
+};
 
 typedef enum {
   INDEXSTORE_SYMBOL_LANG_C = 0,
@@ -289,7 +305,7 @@ typedef enum {
   INDEXSTORE_SYMBOL_LANG_SWIFT = 100,
 } indexstore_symbol_language_t;
 
-typedef enum INDEXSTORE_OPTIONS {
+INDEXSTORE_OPTIONS(indexstore_symbol_role_t, uint64_t) {
   INDEXSTORE_SYMBOL_ROLE_DECLARATION  = 1 << 0,
   INDEXSTORE_SYMBOL_ROLE_DEFINITION   = 1 << 1,
   INDEXSTORE_SYMBOL_ROLE_REFERENCE    = 1 << 2,
@@ -313,7 +329,7 @@ typedef enum INDEXSTORE_OPTIONS {
   INDEXSTORE_SYMBOL_ROLE_REL_CONTAINEDBY = 1 << 16,
   INDEXSTORE_SYMBOL_ROLE_REL_IBTYPEOF    = 1 << 17,
   INDEXSTORE_SYMBOL_ROLE_REL_SPECIALIZATIONOF = 1 << 18,
-} indexstore_symbol_role_t;
+};
 
 INDEXSTORE_PUBLIC indexstore_symbol_language_t
 indexstore_symbol_get_language(indexstore_symbol_t);
--- a/include/indexstore/indexstore.h
+++ b/include/indexstore/indexstore.h
@@ -85,25 +85,23 @@
 #endif
 
 #if __has_attribute(flag_enum)
-# define _INDEXSTORE_FLAG_ENUM __attribute__((flag_enum))
+# define INDEXSTORE_FLAG_ENUM_ATTR __attribute__((flag_enum))
 #else
-# define _INDEXSTORE_FLAG_ENUM
+# define INDEXSTORE_FLAG_ENUM_ATTR
 #endif
 
 #if __has_attribute(enum_extensibility)
-# define _INDEXSTORE_OPEN_ENUM __attribute__((enum_extensibility(open)))
+# define INDEXSTORE_OPEN_ENUM_ATTR __attribute__((enum_extensibility(open)))
 #else
-# define _INDEXSTORE_OPEN_ENUM
+# define INDEXSTORE_OPEN_ENUM_ATTR
 #endif
 
-#define _INDEXSTORE_OPTIONS_ATTRS _INDEXSTORE_OPEN_ENUM _INDEXSTORE_FLAG_ENUM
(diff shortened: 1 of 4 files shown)


kastiglione:
~This change has some slight overlap with #256 and #281, both contain the same `#ifndef __has_attribute` handling. Once one merges, I can update the other.~

kastiglione:
cc @benlangmuir

kastiglione:
Friendly ping @benlangmuir

benlangmuir:
Sorry this got lost.  I don't think this change is safe. Using the enum types as  the return value changes the ABI (they will become 32bit int), and could cause it to silently change again once we add enough cases that they need 64bits of space again.  In C++ we could specify the underlying enum type, but that isn't standard in C.  While clang can support it as a C extension, I'm not sure we want to rely on that for other compilers

kastiglione:
Can the `flag_enum` attribute change go in, even if the return types are left unchanged?

No to push back on anything you said, I am curious about two parts:

1. Are you able to share more about the ABI requirements? I wouldn't have guessed that there are clients of libIndexStore that don't get rebuilt along with updates.
2. What are the downsides to relying on extensions to make the enum 64bit?

Thank you, as always, for reviewing

benlangmuir:
> Can the flag_enum attribute change go in, even if the return types are left unchanged?

Yeah, that should be fine.

> Are you able to share more about the ABI requirements? I wouldn't have guessed that there are clients of libIndexStore that don't get rebuilt along with updates.

The indexer (IndexStoreDB, or Xcode's internal indexer) load this library with dlopen from the toolchain, and need to support multiple toolchain versions at the same time. So this library needs to be ABI stable - i.e. we cannot remove existing symbols or change their signatures, although it's fine to add new functions.   We could potentially break the ABI by updating the major version and updating all the clients to handle the new ABI.  But our clients would need to be able to handle both the old and new ABIs for a while (probably 1 year or more) in order to transition. So we'd need a really good reason to do that.

> What are the downsides to relying on extensions to make the enum 64bit?

It just makes the library less portable.  We're in the process of upstreaming this library to llvm.org, and upstream clang is built by several different compilers for different platforms.  If this extension is available in all those compilers, and doesn't produce any warnings in the default setup, maybe it would be fine to use.

kastiglione:
Following up on the 64bit point, you had said:

> and could cause it to silently change again once we add enough cases that they need 64bits of space again.

If enough cases are added to push it over 32bit, then compiler extensions will be required to make the enum 64bits. Is this enough motivation to go down that path sooner than later? Or do you think it's better to wait until the need forces their use?

benlangmuir:
> If enough cases are added to push it over 32bit, then compiler extensions will be required to make the enum 64bits. Is this enough motivation to go down that path sooner than later? Or do you think it's better to wait until the need forces their use?

Better to wait, I think.  Even if we had large enough values, we don't need to specify an explicit size unless the compiler doesn't support implicitly widening it.

kastiglione:
TIL that clang implicitly widens the size of an `enum` to `long` if its member values don't fit `int`.

kastiglione:
Would it work if we added `_reserved` members to these enums, with a value such as `1L << 63` to force these types to be 64bit now?

benlangmuir:
I took a look and that works on most compilers https://godbolt.org/z/6HpZjW, but it fails on MSVC when building in C mode (in C++ it's fine).

If you can find a workaround for msvc I'd be okay with taking the change.

kastiglione:
Thanks for investigating!

I'm not seeing anything that would make it work for MSVC C. Maybe Apple does not need Windows need ABI compatibility at this time, like it does elsewhere?

If MSVC C does not support larger enums – I'm not finding any extensions to allow it, then these option set types would never support more than 32 members without breaking MSVC C.

benlangmuir:
> I'm not seeing anything that would make it work for MSVC C. Maybe Apple does not need Windows need ABI compatibility at this time, like it does elsewhere?

😞 that's unfortunate.  I don't think we want to have an unstable ABI, since that makes the library unsafe to use in a lot of context. We could theoretically bump the major version at the point we actually need 64 bits, but that would be a big disruption.

What about using a typedef of uint64_t on platforms that don't support this?  For adding cases to the enum it would be a mess once we get to 33 bits, but we could define global constants if necessary.

For example, what if we did something like the `NS_OPTIONS` macro, which checks for C++11 as well as clang's extension to allow fixed size enums outside C++, and if those aren't available it uses a typedef?  That way we get the behaviour we want in clang and swift and C++11, and other cases fallback to a typedef, which is ABI-compatible.

kastiglione:
> What about using a `typedef` of `uint64_t` on platforms that don't support this?

I was just thinking exactly this too. This plan sounds good to me. 👍

kastiglione:
I've updated the PR to add `INDEXSTORE_OPTIONS`, which is just like `{NS,CF}_OPTIONS`. The bit field types (`indexstore_symbol_property_t` and `indexstore_symbol_role_t`) will now be declared as a sized `enum` when used in C++ or ObjC contexts, and otherwise these types will be `typedef`ed accordingly.

benlangmuir (commented):
To match NS_OPTIONS, I think we want to pull the `typedef` out of the macro.

kastiglione:
The reason I had the `typedef` in the macro is because C++ doesn't need a `typedef`, since `enum` types are usable by name. Also, the `typedef` results in a compilation error with c++98 (see below), but does c++98 matter?

For contrast, without typedef:

```
enum NAME : uint64_t {
    ...
};
```

With typedef:

```
typedef enum NAME : uint64_t NAME;
enum NAME : uint64_t {
    ...
};
```

In ObjC, the `typedef` lets the enum be referenced directly by name (`foo_t` vs `enum foo_t`), but C++ already allows enums to be referenced without the keyword. I thought it was weird to introduce an unnecessary `typedef`.

In the "with `typedef`" case, c++98 results in a compilation error:

```
error: definition of type 'foo_t' conflicts with typedef of the same name
```

This could be worked around by prepending or appending an underscore to the name, if compiled under c++98.

include/indexstore/indexstore.h
```diff
@@ -84,6 +84,28 @@
 # define INDEXSTORE_NOESCAPE
 #endif
 
+#if __has_attribute(flag_enum)
+# define _INDEXSTORE_FLAG_ENUM __attribute__((flag_enum))
```
benlangmuir: `_[A-Z]` is a reserved identifier in C.  How about `INDEXSTORE_FLAG_ENUM_ATTR`, `INDEXSTORE_OPEN_ENUM_ATTR` and `INDEXSTORE_OPTIONS_ATTRS`?

include/indexstore/indexstore.h
```diff
@@ -84,6 +84,28 @@
...
+#if __has_attribute(enum_extensibility)
+# define _INDEXSTORE_OPEN_ENUM __attribute__((enum_extensibility(open)))
+#else
+# define _INDEXSTORE_OPEN_ENUM
+#endif
+
+#define _INDEXSTORE_OPTIONS_ATTRS _INDEXSTORE_OPEN_ENUM _INDEXSTORE_FLAG_ENUM
+
+#if __has_extension(cxx_strong_enums)
+# define INDEXSTORE_OPTIONS(_name, _type) enum _INDEXSTORE_OPTIONS_ATTRS _name : _type
```
benlangmuir: NS_OPTIONS is the other order (_type, _name).

benlangmuir (commented):
Huh, I see NS_OPTIONS  in C++ is actually using a typedef over the underlying type and then declaring a separate anonymous enum.  Oh well.

include/indexstore/indexstore.h
kastiglione: Does this apply to for macros too? I was following suit with `CFAvailability.h` (ex `__CF_ENUM_ATTRIBUTES`).

kastiglione:
> Huh, I see NS_OPTIONS in C++ is actually using a typedef over the underlying type and then declaring a separate anonymous enum.

Yeah. Which way do you prefer this to be implemented here? I think for my Swift bridging motivations, the C++ case doesn't matter so I'm happy to follow `NS_OPTIONS` even though it seems less ideal.

include/indexstore/indexstore.h
kastiglione: I'll change it to that. Fwiw, I had it `(_name, _type)` to match the order of `enum` syntax. Interestingly, `OS_ENUM` does the `(_name, _type)` order too ¯\\\_(ツ)_/¯

include/indexstore/indexstore.h
benlangmuir: Yes it does.  `CFAvailability.h` is effectively part of the system (at least on Apple platforms), so it gets to use reserved names.  Same for e.g. the libc or libc++.  While this header is in the compiler repo, it's probably best to treat it more like a normal library header and avoid reserved identifiers.

benlangmuir:
I would mimic NS_OPTIONS as long as it does what you need it to.  If we get C++ import support and need to change it in the future, that should be fine.

include/indexstore/indexstore.h
kastiglione: 👍 Good to know.

benlangmuir:
@swift-ci please test

kastiglione:
Does swift-ci post a comment if there are failures/issues?

benlangmuir:
Oh, I think it's disabled on this branch.  @shahmishal @hyp where can we test and then merge this change? I guess it should go into upstream-with-swift, but is that testable?  Do we have a branch that feeds into stable other than 5.1 that this can go on?

kastiglione:
I'm happy to rebase this. In my previous related change, I did two PRs, one that targeted `upstream-with-swift` and another that targeted `swift-5.1-branch`.

benlangmuir:
@hyp tells me we want:
* upstream-with-swift
* apple/stable/20190619

swift-5.1 is locked down at this point

kastiglione:
I've changed this PR to target `upstream-with-swift`. I'll open a PR to target `apple/stable/20190619`.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

kastiglione:
Seems like the PR into `apple/stable/20190619 ` can be closed https://github.com/apple/swift-clang/pull/352. Is this still good? Should I open PRs into any other branches?

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

kastiglione:
would it help to rebase?

kastiglione:
I rebased.

benlangmuir:
I don't see how that would help, but we can try.  The failures looked unrelated to me.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
Actually, after talking with Mishal, we decided to just merge this.

kastiglione:
@benlangmuir thanks for helping me see this through

PR #290, opened 2019-04-03
CodeGen: correct merge mixup

compnerd:
`EmitRuntimeCall` now requires a `llvm::FunctionCallee` which is being generated
and returned as is.  Fix the signature.


compnerd:
CC: @bob-wilson @dexonsmith

bob-wilson:
Slava applied this change in fe297b2db47fd650a48f3ac70cd9507a0c5abce4

PR #291, opened 2019-04-04
[IndexStore] Enable blocks support on Linux builds

kastiglione:
This enables block support for Linux builds of `IndexStore`. The motivation is to have the various `_apply` functions be present (ex `indexstore_store_units_apply`). Without this, only the function pointer variants are available (ex `indexstore_store_units_apply_f`).

### DIFF
--- a/tools/IndexStore/CMakeLists.txt
+++ b/tools/IndexStore/CMakeLists.txt
@@ -49,6 +49,10 @@ if(APPLE)
 
   set_property(TARGET IndexStore APPEND_STRING PROPERTY
                LINK_FLAGS ${INDEXSTORE_LINK_FLAGS})
+else()
+  # Blocks are off by default on non-darwin platforms.
+  target_compile_options(IndexStore PRIVATE -fblocks)
+  target_link_libraries(IndexStore PRIVATE BlocksRuntime)
 endif()
 
 if (LLVM_INSTALL_TOOLCHAIN_ONLY)


kastiglione:
Note that the Swift 5 Linux release includes a libIndexStore.so that is missing those symbols.

kastiglione:
cc @benlangmuir

tools/IndexStore/CMakeLists.txt
```diff
@@ -49,6 +49,10 @@ if(APPLE)
 
   set_property(TARGET IndexStore APPEND_STRING PROPERTY
                LINK_FLAGS ${INDEXSTORE_LINK_FLAGS})
+else()
+  # Blocks are off by default on non-darwin platforms.
+  target_compile_options(IndexStore PRIVATE -fblocks)
```
compnerd, line 54: This can be hoisted out of the condition as it is safe to apply `-fblocks` when building for Darwin.  However, please ensure that you do not apply this when the C++ compiler is not clang.  The Windows builds use cl to build the toolchain as it is much faster and better.

tools/IndexStore/CMakeLists.txt
```diff
@@ -49,6 +49,10 @@ if(APPLE)
...
+  target_link_libraries(IndexStore PRIVATE BlocksRuntime)
```
compnerd, line 55: Ugh, this is really annoying.  The `BlocksRuntime` really should come from libdispatch, as that is the canonical one that is used elsewhere in the swift project.  I don't know how to break this cycle.

PR #294, opened 2019-04-10
Fixing swift-clang tests that fail due to missing PrintingPolicy init in ctor.

plotfi:
The following tests that were failing were:

    Clang :: OpenMP/for_simd_ast_print.cpp
    Clang :: OpenMP/simd_ast_print.cpp
    Clang-Unit :: AST/./ASTTests/DeclPrinter.TestFunctionDecl4
    Clang-Unit :: AST/./ASTTests/DeclPrinter.TestFunctionDecl5

They were failing because SupressStorageClassSpecifiers not being properly
initialized in PrintingPolicy in include/clang/AST/PrettyPrinter.h and when it
was used in lib/AST/DeclPrinter.cpp "static" was not being printed:

```
  if (!Policy.SuppressSpecifiers) {
    if (!Policy.SupressStorageClassSpecifiers) {
      switch (D->getStorageClass()) {
      case SC_None: break;
      case SC_Extern: Out << "extern "; break;
      case SC_Static: Out << "static "; break;
```

These tests pass in upstream clang because SupressStorageClassSpecifiers was
never pushed upstream.

### DIFF
--- a/include/clang/AST/PrettyPrinter.h
+++ b/include/clang/AST/PrettyPrinter.h
@@ -38,6 +38,7 @@ struct PrintingPolicy {
   /// Create a default printing policy for the specified language.
   PrintingPolicy(const LangOptions &LO)
       : Indentation(2), SuppressSpecifiers(false),
+        SupressStorageClassSpecifiers(false),
         SuppressTagKeyword(LO.CPlusPlus), IncludeTagDefinition(false),
         SuppressScope(false), SuppressUnwrittenScope(false),
         SuppressInitializers(false), ConstantArraySizeAsWritten(false),


compnerd:
CC: @bob-wilson @gottesmm @dexonsmith @hyp

plotfi:
I have to also note, this was failing on swift-clang and not on llvm.org's clang because 9890adfbee8 was never upstreamed. This seems kinda odd. Why was this not upstreamed??

gottesmm:
@plotfi it was probably an oversight. @hyp your thoughts?

gottesmm:
I would really like us to have a bot on ci.swift.org that checks that upstream-with-swift/master can pass all llvm tests. @shahmishal

shahmishal:
@gottesmm Can you please file a ticket on bugs.swift.org?

compnerd:
@swift-ci please test and merge

compnerd:
Oh, bleh, guess there aren't bots for this branch yet.  Well, tested this locally, and this is improving things, so lets get this merged.

PR #295, opened 2019-04-10
Fixing SemaObjC tests that take -x objective-c++ with -stdlib=libc++

plotfi:
The following get fixed with this change (on Linux anyways):

    Clang :: SemaObjC/class-stub-attr-unsupported.m
    Clang :: SemaObjC/class-stub-attr.m


compnerd:
CC: @dexonsmith @slavapestov

PR #296, opened 2019-04-11
Fixing clang-refactor-test tests on Linux by adding -fobjc-runtime=ios-5.0

plotfi:
On Linux, without -fobjc-runtime=ios-5.0 the following:

```
@implementation NoNeedForImplementationTUs {
  int _p2;
}
```

will get AST synthesized without:

```
`-ObjCPropertyImplDecl 0x7fad8c077070 <<invalid sloc>, line:21:15> <invalid sloc> p2 synthesize
  |-ObjCProperty 0x7fad8c069b68 'p2'
  `-ObjCIvar 0x7fad8c069a68 '_p2' 'int'
```

In this case, the refactoring wont find the decl in the @implementation.


compnerd:
CC: @hyp

PR #298, opened 2019-04-15
Basic: adjust for overflow in diagnostics

compnerd:
Recent upstream changes has caused the number of diagnostics to overflow.
Increase the count.

### DIFF
--- a/include/clang/Basic/DiagnosticIDs.h
+++ b/include/clang/Basic/DiagnosticIDs.h
@@ -36,7 +36,7 @@ namespace clang {
       DIAG_SIZE_AST           =  150,
       DIAG_SIZE_COMMENT       =  100,
       DIAG_SIZE_CROSSTU       =  100,
-      DIAG_SIZE_SEMA          = 3500,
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @bob-wilson @dexonsmith @gottesmm

gottesmm:
@jrose-apple would be good to look at this. On a more meta level, why only increase by 3? Why not bump it by 100 or something?

compnerd:
@gottesmm - thats what the current need is which is why.  I'm not tied to 3 :-)

compnerd:
CC: @atrick

bob-wilson:
Bumping it by 3 (or however many more diagnostics we have in swift-clang) seems OK to me. The limit on trunk will get bumped soon, and presumably that will go up by a significant amount. E.G., if the trunk limit goes from 3500 to 3600, the swift-clang version can go from 3503 to 3603 (or even just leave it at 3600 and repeat this process when/if we get close to that limit).

gottesmm:
@bob-wilson I agree it is reasonable. +1 from me.

PR #300, opened 2019-04-16
[FileSystemStatCache] Return std::error_code from stat cache methods

harlanhaskins:
(cherry-picking from https://reviews.llvm.org/D60735)

Previously, we would return true/false signifying if the cache/lookup
succeeded or failed. Instead, provide clients with the underlying error
that was thrown while attempting to look up in the cache.

Since `clang::FileManager` doesn't make use of this information, it discards the
error that's received and casts away to bool.

This change is NFC.


harlanhaskins:
@swift-ci please test

PR #302, opened 2019-04-19
Basic: increase diagnostics again

compnerd:
Until the local diagnostics are upstreamed, we need to track the number of
diagnostics that we are adding.  For now, just bump up the count.


compnerd:
CC: @bob-wilson @dexonsmith @atrick

dexonsmith (approved):
LGTM.

PR #303, opened 2019-04-23
Diagnostics count

compnerd:


compnerd:
CC: @bob-wilson @dexonsmith

bob-wilson:
I see 30 additional entries in DiagnosticSemaKinds.td compared to trunk. I think you should bump the limit to be 3530.

compnerd:
Sounds good to me!

PR #305, opened 2019-04-29
Revert "[Modules] Make header inclusion order from umbrella dirs deterministic"

jrose-apple:
Reverts apple/swift-clang#304. It's causing problems on CI still.


lib/Frontend/FrontendAction.cpp
```diff
@@ -375,20 +374,13 @@ static std::error_code collectModuleHeaderIncludes(
            ++It)
         llvm::sys::path::append(RelativeHeader, *It);
 
-      Headers.push_back(std::make_pair(RelativeHeader.str(), Header));
```
jrose-apple, line 378: Ah, maybe this `str()` was important.

PR #306, opened 2019-04-29
Swift 5.1 branch sort umbrellas retry

bcardosolopes:


bcardosolopes:
I've tested this in https://github.com/apple/swift/pull/24380, multiple times (the only failure is an unrelated problem that's happening in the bots). Will attempt to merge this again, hopefully it will work this time.

PR #312, opened 2019-05-09
[IndexDataStore] Sort unit names using std::sort()

rintaro:
std::string is not POD.


rintaro:
@swift-ci please test

PR #330, opened 2019-06-13
[tensorflow branch] add InMemoryOutputFileSystem for module caches

marcrasi:
Context about the problem: We'd like to run SourceKit on a server where we're not allowed to write to the filesystem. The ClangImporter writes module caches to the filesystem, so we need some way to stop that from happening.

This PR solves the problem by defining an "InMemoryOutputFileSystem" that collects the built modules into memory and serves them through a `llvm::vfs::FileSystem` interface.

Corresponding swift repo PR (which includes a test) is here: https://github.com/apple/swift/pull/25438

This may only be a temporary solution, but everything else I tried was quite hard to get working (explicitly building required clang modules before invoking SourceKit, so that SourceKit doesn't have to build the modules into the cache), so I'd like to try this for a while.

### DIFF
--- a/include/clang/Basic/InMemoryOutputFileSystem.h
+++ b/include/clang/Basic/InMemoryOutputFileSystem.h
@@ -1,0 +1,106 @@
+//===-- InMemoryOutputFileSystem.h - Collects outputs in memory -*- C++ -*-===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_CLANG_BASIC_INMEMORYOUTPUTFILESYSTEM_H_
+#define LLVM_CLANG_BASIC_INMEMORYOUTPUTFILESYSTEM_H_
+
+#include <memory>
+#include "llvm/ADT/IntrusiveRefCntPtr.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringMap.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Mutex.h"
+#include "llvm/Support/MutexGuard.h"
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/VirtualFileSystem.h"
+
+namespace clang {
+
+/// Collects output files in memory, and provides a `llvm::vfs::FileSystem`
+/// interface for accessing those files.
+///
+/// This class is threadsafe. Unsynchronized calls from multiple threads will
+/// not corrupt the internal state, and operations occur atomically and
+/// sequentially consistently from the point of view of all threads.
+class InMemoryOutputFileSystem : public llvm::vfs::FileSystem {
+ public:
+  InMemoryOutputFileSystem() : OutputFiles(new llvm::vfs::InMemoryFileSystem())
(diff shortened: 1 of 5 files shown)


include/clang/Frontend/CompilerInstance.h
```diff
@@ -412,6 +417,14 @@ class CompilerInstance : public ModuleLoader {
     VirtualFileSystem = std::move(FS);
   }
 
+  IntrusiveRefCntPtr<clang::InMemoryOutputFileSystem> getInMemoryOutputFileSystem() const {
```
burmako, line 427: Judging by the comments above and below, perhaps it would be appropriate to say:

```
  /// }
  /// @name In-Memory Output File System
  /// {
```

include/clang/Basic/InMemoryOutputFileSystem.h
```diff
@@ -0,0 +1,104 @@
...
+  /// eventually appear on the `llvm::vfs::FileSystem` interface.
+  /// `InMemoryOutputFileSystem` owns the buffer, which will not be released
+  /// until `DeleteTemporaryFile` or `FinalizeTemporaryFile` is called.
+  /// \param OutputPath the path of the file that may eventually be created.
+  /// \param TemporaryPath must be non-null. Pointee will be set to a unique
+  //         string identifying this particular temporary buffer.
+  //  \returns A stream that can be used to write to the buffer.
+  std::unique_ptr<llvm::raw_pwrite_stream> CreateTemporaryBuffer(
+      llvm::StringRef OutputPath,
+      std::string *TemporaryPath);
```
burmako, line 47: Should this also take a pointer to a StringRef?

include/clang/Basic/InMemoryOutputFileSystem.h
```diff
@@ -0,0 +1,104 @@
...
+
+  /// Releases the buffer undelying the temporary file.
```
burmako:
```suggestion
  /// Releases the buffer underlying the temporary file.
```

include/clang/Basic/InMemoryOutputFileSystem.h
```diff
@@ -0,0 +1,104 @@
...
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/VirtualFileSystem.h"
+
+namespace clang {
+
+/// Collects output files in memory, and provides a `llvm::vfs::FileSystem`
+/// interface for accessing those files.
+///
+/// This class is threadsafe.
```
burmako: Could you elaborate on what thread-safety means in this context?

burmako (approved):

include/clang/Basic/InMemoryOutputFileSystem.h
marcrasi: done

include/clang/Frontend/CompilerInstance.h
marcrasi: done

include/clang/Basic/InMemoryOutputFileSystem.h
marcrasi: done

include/clang/Basic/InMemoryOutputFileSystem.h
marcrasi: No, a StringRef is an immutable reference to a string, so this wouldn't be able to set the string if it took a StringRef or a pointer to a StringRef.

PR #333, opened 2019-06-21
[Frontend] Respect SkipUnusedModuleMaps in DependencyCollector

harlanhaskins:
Previously, if SkipUnusedModuleMaps was passed, the DependencyCollector
would still add unused module maps as a dependency. Instead, it should
either add everything or only add those which are actually loaded.

rdar://51713745


bcardosolopes:
It sounds fine as an extra argument given that this is built on top of something that might change when upstreaming. @harlanhaskins can you add a clang testcase for this?

PR #334, opened 2019-06-22
Basic: correct signature for `__iso_volatile_{load,store}32`

compnerd:
Microsoft provides an extension which performs synchronised load/store.
Because the environment is a LLP64 environment, `long` and `int` are
interchangeable.  We need to loosen the signature to permit both to be
accepted.


compnerd:
CC: @dexonsmith @bob-wilson

compnerd:
A note: this is not a cherry-pick and does not have test cases for a reason.  The branch is now too far out of sync from upstream to do a single cherry-pick for this.  I will be doing a separate change upstream for fixing the issue with test cases.  That wont backport cleanly without cherry-picking more changes and I think that it is more risky to do that than to do this special case one off for the 5.1 branch.  This is needed for supporting Windows ARM64 as the C++ runtime uses this for synchronization which is needed for the side table refcounting.  I could add the test cases, but that adds additional divergence, and we will pick up the test cases when we move to the next branch point.

compnerd:
FTR, upstream SVN r364137 is the same change with tests.

compnerd:
Oh, and a last note: this is very low risk as this impacts only the targets which are enabling MS compatibility via `-fms-compatibility`, which is by default only the Windows targets.

compnerd:
Ugh, this does cause some regressions :/  I've reverted it in SVN r364147 while I try to find a better solution.

PR #336, opened 2019-06-25
[5.1] [FileManager] Revert r347205 to avoid PCH file-descriptor leak.

jrose-apple:
Cherry-pick of https://reviews.llvm.org/D57165 (from January!) to fix a file descriptor leak. See there for details.

rdar://problem/49785385

### DIFF
--- a/lib/Basic/FileManager.cpp
+++ b/lib/Basic/FileManager.cpp
@@ -189,20 +189,14 @@ const FileEntry *FileManager::getFile(StringRef Filename, bool openFile,
       *SeenFileEntries.insert(std::make_pair(Filename, nullptr)).first;
 
   // See if there is already an entry in the map.
-  if (NamedFileEnt.second) {
-    if (NamedFileEnt.second == NON_EXISTENT_FILE)
-      return nullptr;
-    // Entry exists: return it *unless* it wasn't opened and open is requested.
-    if (!(NamedFileEnt.second->DeferredOpen && openFile))
-      return NamedFileEnt.second;
-    // We previously stat()ed the file, but didn't open it: do that below.
-    // FIXME: the below does other redundant work too (stats the dir and file).
-  } else {
-    // By default, initialize it to invalid.
-    NamedFileEnt.second = NON_EXISTENT_FILE;
-  }
+  if (NamedFileEnt.second)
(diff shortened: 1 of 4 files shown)


jrose-apple:
@swift-ci Please smoke test

jrose-apple:
**Explanation**: A change in open-source LLVM to fix a bug where opening file descriptors *wouldn't* be cached ended up creating a bug where file descriptors wouldn't be *closed,* which is much worse. This change reverts that original fix, and has been in open-source LLVM since January.

**Scope**: Affects many uses of Clang's `FileManager`.

**Risk**: Low: public LLVM has been living on this, and the revert brings us back to the behavior in Swift 5.0.

**Testing**: The revert includes a new regression test, and I also verified that this fixes the issue with a precompiled bridging header in rdar://problem/49785385.

**Reviewer**: Originally reviewed by Nico Weber.

jrose-apple:
@swift-ci Please test

shahmishal:
@swift-ci test

lib/Basic/FileManager.cpp
```diff
@@ -295,11 +279,13 @@ const FileEntry *FileManager::getFile(StringRef Filename, bool openFile,
   UFE.Dir     = DirInfo;
   UFE.UID     = NextFileUID++;
   UFE.UniqueID = Status.getUniqueID();
-  UFE.IsNamedPipe = Status.getType() == llvm::sys::fs::file_type::fifo_file;
```
vsapsai: I think you need to keep this line, though I haven't tested it. Last time it was changed in [a different commit](https://github.com/apple/swift-clang/commit/e6138c2236eeb616c2f2b9774be0640fb0e026db#diff-0025af005307891b5429b6a834823d5eR295), not the one you are reverting.

vsapsai (changes requested):
`UFE.IsNamedPipe` change (commented inline) looks suspicious, the rest looks good.

lib/Basic/FileManager.cpp
jrose-apple: Oops. Bad merge resolution, thank you for catching this!

jrose-apple:
@swift-ci test

vsapsai (approved):

PR #338, opened 2019-06-27
[DependencyCollector] Add optional reproducer callback

JDevlieghere:
The clang importer in Swift uses its own subclass of the clang
dependency collector. Because it is completely hidden in the clang
importer's implementation, we cannot subclass it from LLDB.

This is an issue for reproducers, because LLDB needs to be informed of
all the files used by clang through the Swift compiler. To solve this, I
added the ability to register a custom callback in the dependency
collector. LLDB can use this callback to be notified, without needing
access to the dependency collector's implementation.

### DIFF
--- a/include/clang/Frontend/Utils.h
+++ b/include/clang/Frontend/Utils.h
@@ -101,6 +101,10 @@ public:
   /// Return true if system files should be passed to sawDependency().
   virtual bool needSystemDependencies() { return false; }
 
+  void registerCallback(std::function<void(StringRef)> C) {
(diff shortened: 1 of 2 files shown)


hyp:
Can this go upstream?

vsapsai:
I want to check more thoroughly potential lifetime issues.

Are there other approaches you've considered? So far this extra callback to me looks awkward and I'm curious if we can do better.

JDevlieghere:
Closing this because I plan a different approach.

PR #339, opened 2019-06-27
[5.1] [FileManager] fillRealPathName even if we aren't opening the file

jrose-apple:
Follow-up to #336 to fix issues in clangd. Credit to @jkorous-apple for noticing that he'd fixed this back when it happened in upstream LLVM.

rdar://problem/52272757

### DIFF
--- a/lib/Basic/FileManager.cpp
+++ b/lib/Basic/FileManager.cpp
@@ -286,6 +286,9 @@ const FileEntry *FileManager::getFile(StringRef Filename, bool openFile,
   if (UFE.File) {
     if (auto PathName = UFE.File->getName())
       fillRealPathName(&UFE, *PathName);
+  } else if (!openFile) {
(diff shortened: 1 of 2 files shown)


jrose-apple:
**Explanation**: The change I recently cherry-picked to fix a bug where file descriptors wouldn't be closed…turned out to reintroduce a bug where files that were *not* opened weren't getting a realpath set. This cherry-picks the fix for that.

**Scope**: Affects users of Clang's FileManager that depend on the realpath of a file that's not opened.

**Risk**: Low; public LLVM has been living on this for a while. I'm unlikely to have made the same mistake again because I checked all the other changes to Clang's FileManager since then.

**Testing**: Includes a new regression test for the condition clangd is relying on.

**Reviewer**: @hyp (patches by @jkorous-apple)

hyp (approved):
LGTM

PR #340, opened 2019-07-11
tweak DirectoryWatcher to use LINK_LIBS instead of target_link_libraries.

pschuh:
This is the root cause of the initial error in master-next build problems.

See https://ci.swift.org/view/swift-master-next/job/oss-swift-incremental-RA-linux-ubuntu-16_04-master-next/11345/console for example.

### DIFF
--- a/lib/DirectoryWatcher/CMakeLists.txt
+++ b/lib/DirectoryWatcher/CMakeLists.txt
@@ -22,6 +22,6 @@ endif()
 
 add_clang_library(clangDirectoryWatcher
   ${DIRECTORY_WATCHER_SOURCES}
+  LINK_LIBS
+  ${DIRECTORY_WATCHER_LINK_LIBS}
   )
-
-target_link_libraries(clangDirectoryWatcher PRIVATE ${DIRECTORY_WATCHER_LINK_LIBS})


compnerd:
These should be largely the same.  How does changing this change the behaviour?  This will be the only case where we do this - and is definitely something which really goes against best practices for CMake. (CC: @llvm-beanz)

jkorous-apple:
Thanks. LGTM. Let's fix this now. We can use more elegant solution later.

llvm-beanz:
This issue was discovered on llvm.org and a patch to resolve it landed earlier today.

llvm-beanz:
See: https://reviews.llvm.org/D64579

pschuh:
The precise nature of the bug is that this adds $<LINK_ONLY:-lpthread> to:

lib/cmake/clang/ClangTargets.cmake
```
set_target_properties(clangDirectoryWatcher PROPERTIES
  INTERFACE_LINK_LIBRARIES "LLVMSupport;\$<LINK_ONLY:-lpthread>"
)
```
This will then show up in build.ninja as an unescaped: $<LINK_ONLY:-lpthread>

@llvm-beanz Note that this particular file is swift-clang specific. Is your solution generic for all libs?

llvm-beanz:
What target is that showing up on? My guess is clang_shared, which is what this patch addresses. `INTERFACE_LINK_LIBRARIES` should support generator expressions just fine. The problem is that CMake pre-3.13 doesn't handle expanding nested generator expressions, which is what clang_shared was generating.

compnerd:
I believe that the right thing to do here is to just always link against the thread libraries:

```
target_link_libraries(clangDirectoryWatcher PUBLIC
  ${CMAKE_THREAD_LIBS_INIT})
```

pschuh:
@llvm-beanz Confirmed your fix works for this.

PR #342, opened 2019-07-12
Adding sys::fs::convertFDToNativeFile(FD), to address llvm.org/D63452

plotfi:
IndexUnitReader.cpp is not in upstream llvm. It was using an older API.


compnerd:
CC: @jkorous-apple @hyp @dexonsmith

plotfi:
ping

jkorous-apple (approved):
LGTM. Thanks.

PR #343, opened 2019-07-12
DirectoryWatcher for Windows stubs (to fix build break).

plotfi:
This is just a code skeleton for DirectoryWatcher-windows.cpp so the
build on Windows stops breaking.


compnerd:
CC: @jkorous-apple @hyp

jkorous-apple (approved):
LGTM

compnerd:
Yeah, lets get the build going and then we can implement the functionality.

PR #348, opened 2019-08-10
CodeGen: ensure 8-byte aligned String Swift CF ABI

compnerd:
CFStrings should be 8-byte aligned when built for the Swift CF runtime
ABI as the atomic CF info field must be properly aligned.  This is a
problem on 32-bit platforms which would give the structure 4-byte
alignment rather than 8-byte alignment.

git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@368471 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit ee5531b2162e8d696928cbd48f58d92740c1015e)

### DIFF
--- a/lib/CodeGen/CodeGenModule.cpp
+++ b/lib/CodeGen/CodeGenModule.cpp
@@ -4360,7 +4360,10 @@ CodeGenModule::GetAddrOfConstantCFString(const StringLiteral *Literal) {
   }
   Fields.addInt(LengthTy, StringLength);
 
-  CharUnits Alignment = getPointerAlign();
+  // Swift ABI requires 8-byte alignment to ensure that the _Atomic(uint64_t) is
+  // properly aligned on 32-bit platforms.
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @dexonsmith @millenomi 

This is needed to repair Foundation on non-Darwin 32-bit targets.  Since this is controlled by the `-fcf-runtime-abi=` flag, with the default being `ObjC`, this has no impact on the Darwin path as Darwin targets do not currently use the Swift runtime ABI (due to historical accidents).  Additionally, this patch *only* changes things on 32-bit targets (watchOS is excluded as it is a Darwin target), as 64-bit targets would use `sizeof(void *)` as the alignment (which would be 8-bytes).  Even there, it is increasing the alignment (4 -> 8) and should be safe for interoperability (even though there is not ABI guarantees yet on non-Darwin targets).

dexonsmith (approved):
LGTM.

PR #349, opened 2019-08-14
Expose virtual method lookup and this ptr adjustment.

pschuh:
Companion PR: https://github.com/apple/swift/pull/26658

### DIFF
--- a/include/clang/CodeGen/CodeGenABITypes.h
+++ b/include/clang/CodeGen/CodeGenABITypes.h
@@ -66,6 +66,9 @@ const CGFunctionInfo &arrangeCXXMethodType(CodeGenModule &CGM,
                                            const FunctionProtoType *FTP,
                                            const CXXMethodDecl *MD);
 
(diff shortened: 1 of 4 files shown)


hyp:
This change should land on llvm.org/master first, unless it depends on code on GitHub.com/apple/swift-clang. If it does, it should be committed to `upstream-with-swift` branch first, not `stable`, and then cherry-picked to `apple/stable/20190619` (merge window for `swift-5.1-branch` is already closed).

pschuh:
@hyp I think that this can go into master directly. I'll leave this PR until I have this sorted out so it can be used as reference for the other PR. Thanks!

rjmccall:
Please do CC me on the Clang code review, I have thoughts about this but they should be raised in one place.

pschuh:
@rjmccall I added: https://reviews.llvm.org/D66360 but I don't know who to put as the reviewer. Could that just be you?

rjmccall:
Yes, just putting me as the reviewer is fine.

PR #351, opened 2019-08-15
Replace last uses of llvm::index_sequence{,_for}

bulbazord:
After the next merge, swift-clang will fail to compile because of a few straggling uses of llvm::index_sequence and llvm::index_sequence_for. Let's replace them with those available in c++.

cc @dcci @aschwaighofer @compnerd


dcci (approved):
LGTM.

PR #354, opened 2019-09-03
[CrashReproducer] Handle indexing turned on with CLANG_PROJECT_INDEX_PATH.

vsapsai:
Crash reproducer is checking the driver's arguments for
`-index-store-path` and in case of the environment variable, this
argument is absent. So the generated reproducer incorrectly won't run
the indexer.

Fix by checking the arguments of the failed -cc1 command for
`-index-store-path`.

rdar://problem/48830606

### DIFF
--- a/include/clang/Driver/Job.h
+++ b/include/clang/Driver/Job.h
@@ -31,11 +31,12 @@ class Tool;
 struct CrashReportInfo {
   StringRef Filename;
   StringRef VFSPath;
-  StringRef IndexStorePath;
+  bool HaveIndexStorePath;
 
   CrashReportInfo(StringRef Filename, StringRef VFSPath,
-                  StringRef IndexStorePath)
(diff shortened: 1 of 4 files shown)


jkorous-apple:
@swift-ci please test

jkorous-apple:
@vsapsai just wondering - how do we handle other environment variables in reproducers? Strictly speaking this isn't a 100% accurate reproducer (although for all practical purposes it shouldn't matter).

vsapsai:
> @vsapsai just wondering - how do we handle other environment variables in reproducers? Strictly speaking this isn't a 100% accurate reproducer (although for all practical purposes it shouldn't matter).

Good question. Need to double check but my impression is that the reproducer contains `-cc1` options and environment variables are handled by the driver. So all the extra stuff the driver has added should be preserved. But `-index-store-path` is special because we don't preserve its value verbatim but change to a different one in the reproducer directory. That's how it slipped through the cracks.

PR #360, opened 2019-10-03
[DirectoryWatcher] Fix cross-compilation checking fs_event availability.

dcci:
Some platforms have the header but don't have support for fs_event.


dcci:
@swift-ci test

jkorous-apple (commented):
Could you please update the Cmake check in tests too?

https://github.com/apple/swift-clang/blob/stable/unittests/DirectoryWatcher/CMakeLists.txt

PR #361, opened 2019-10-03
[DirectoryWatcher] Fix cross-compilation checking fs_event availability.

dcci:
Some platforms have the header but don't have support for fs_event.

### DIFF
--- a/lib/DirectoryWatcher/DirectoryWatcher.cpp
+++ b/lib/DirectoryWatcher/DirectoryWatcher.cpp
@@ -92,7 +92,11 @@ struct DirectoryScan {
 # define __has_include(x) 0
 #endif
 
-#if __has_include(<CoreServices/CoreServices.h>)
+#if !defined(__is_target_os)
+#define __is_target_os(x) 0
+#endif
+
+#if __is_target_os(macos)
 # include "DirectoryWatcher-mac.inc.h"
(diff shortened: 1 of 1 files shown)


dcci:
@swift-ci test

lib/DirectoryWatcher/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ include(CheckIncludeFiles)
 
 set(LLVM_LINK_COMPONENTS support)
 
+CHECK_SYMBOL_EXISTS("FSEventStreamRef" "CoreServices/CoreServices.h" HAS_FS_EVENTS)
```
akyrtzi: In llvm it is called with the lowercased version, `check_symbol_exists`.
Also the macro will not be automatically be defined, you need to add a `#cmakedefine` in `include/clang/Config/config.h.cmake`.

For an example see how `HAVE_DLOPEN` is checked and defined in the llvm repo.

akyrtzi (changes requested):

dcci:
Followed up in person with Argyrios and decided it's simpler to check the platform instead [at least here, maybe we'll do something different upstream]

jkorous-apple (approved):
LGTM

dcci:
@swift-ci test

dcci:
@swift-ci test

PR #365, opened 2019-10-22
Fix FileManager invalidation issue for file redirect

asuhan:


asuhan:
@marcrasi Not sure what the best process for this fix would be, feel free to suggest a different way.

